PR #5, opened 2023-02-04
fix: automatically choose whether to disable checksum according to NIC offload

mzz2017:


mzz2017:
Close due to manual merging by https://github.com/v2rayA/dae/commit/0d29b6fccc0ac02d799b453c3ac607f32946cca8

PR #10, opened 2023-02-08
feat: support independent tcp4, tcp6,  udp4, udp6 connectivity check

mzz2017:
This is a break change for config file.


cubercsl:
@mzz2017 nb

mzz2017:
Now ipv4/ipv6/UDP/TCP may select different nodes in the group according to the selection policy (for example, min latency).

PR #15, opened 2023-02-12
docs: add ip-forwarding guide

yqlbu:
## Summary

As the title suggests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 <img src="https://github.com/v2rayA/dae/blob/main/logo.png" border="0" width="25%">
 
-***dae***, means goose, is a lightweight and high-performance transparent proxy solution.
+**_dae_**, means goose, is a lightweight and high-performance transparent proxy solution.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -67,19 +67,33 @@ Check them using command like:
 (zcat /proc/config.gz || cat /boot/{config,config-$(uname -r)}) | grep 'CONFIG_DEBUG_INFO_BTF='
 ```
 
+### Enable IP Forwarding
+
+By default, any latest Linux distributions will have IP Forwarding `disabled`. In the case where we need to up a Linux router/gateway or a VPN server or simply a plain dial-in server, then we must need to enable forwarding. Do the followings to have `ip-forwarding` feature enabled.
+
+```shell
+cat >>/etc/sysctl.d/dae.conf<<EOF
```
mzz2017, line 75: User may need `sudo` here.

mzz2017 (changes requested):

README.md
yqlbu: Resolved in [commit](https://github.com/v2rayA/dae/pull/15/commits/1d2aa7d526bf65758735c873a90c0e4fd6202248)

mzz2017: You cannot use `sudo` to redirect output. You can try.

mzz2017: We can use:
```
sudo tee /etc/sysctl.d/dae.conf << 'EOF'
net.ipv4.ip_forward = 1
net.ipv6.conf.all.forwarding = 1
EOF
```

yqlbu: Sure. Let's use the above fix.

mzz2017:
LGTM. Thanks for your contribution~

PR #35, opened 2023-03-14
feat: shorten docker command arguments by leveraging CMD

kunish:
As the title suggests, this PR is to shorten docker command user arguments by adding default `ENTRYPOINT` and 
`CMD` to Dockerfile, which is used in the CI/CD build steps.

Before

```sh
docker exec dae dae reload
```

After

```sh
docker exec dae reload
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,4 +15,6 @@ RUN wget -O /usr/local/share/dae/geosite.dat https://github.com/v2ray/domain-lis
 COPY --from=builder /build/dae /usr/local/bin
 COPY --from=builder /build/install/empty.dae /etc/dae/config.dae
 RUN chmod 0600 /etc/dae/config.dae
-CMD ["dae", "run", "-c", "/etc/dae/config.dae"]
+
+CMD ["dae"]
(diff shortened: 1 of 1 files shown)


mzz2017:
Thanks for your PR!

dae does not listen on port 2023 like dae-wing, thus we have no need to expose it.

And why we need an independent CMD [ ] ?

kunish:
See [this article](https://sharonsahadevan.medium.com/understanding-the-differences-between-docker-cmd-and-docker-entrypoint-when-to-use-each-in-your-d580a7b616f8), which explains in detail the main differences and usecases between these two Dockerfile commands.

In short

> Use CMD to specify the default command to be executed when running a container from the image.

> Use ENTRYPOINT to specify the main command to execute when the container starts.

kunish:
> dae does not listen on port 2023 like dae-wing, thus we have no need to expose it.

My mistake. I was confused with another project we are working on.

mzz2017:
@kunish Thank you for your explanation. I think it has been ready to merge.

kunish:
Just a quick reminder, we need to also change the instructions on [dockerhub README](https://hub.docker.com/r/daeuniverse/dae)

PR #39, opened 2023-03-24
doc: update guide to debian-kernel-upgrade

yqlbu:
## Summary

The previous guide does NOT actually work. I've tested it on a clean Debian 11 install. Hence, I proposed the following draft changes.


mzz2017:
Cool. I'll merge it right now.

PR #44, opened 2023-03-26
doc(example.dae): add node selection by tag

yqlbu:
## Summary

Add an example to select target nodes by tag.


example.dae
```diff
@@ -119,6 +121,8 @@ group {
     group2 {
         # Filter nodes from the global node pool defined by the subscription and node section above.
         filter: subtag(regex: '^my_', another_sub) && !name(keyword: 'ExpireAt:')
+        # Filter nodes from the global node pool defined by tag
+        # filter: name(node1, node2)
```
mzz2017, line 125: Suggest:
```
        # Filter nodes from the global node pool defined by tag
        #filter: name(node1, node2)
```

yqlbu: Resolved.

PR #51, opened 2023-04-01
ci: use dedicated self-hosted runners

yqlbu:
## Summary

The `min_num` runner is set to `13` as shown in the screenshot below. Max capacity is set to `39` for now. Let's have a try.

image


yqlbu:
Will reopen once networking issue gets fixed.

PR #70, opened 2023-05-01
fix: sniffed domain with brackets

mzz2017:
## Motivation

`curl http://[2606:4700:20::681a:d1f]:80` will fail in `domain++` mode.

```
level=warning msg="handleConn: failed to dial [2606:4700:20::681a:d1f]:80: SplitHostPort: address [[2606:4700:20::681a:d1f]]:80: missing port in address"
```

## Modification

Remove duplicated brackets from sniffed ipv6-like domain.

## Test

The problem disappeared after the fix.

### DIFF
--- a/control/control_plane.go
+++ b/control/control_plane.go
@@ -18,6 +18,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/bits-and-blooms/bloom/v3"
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/rlimit"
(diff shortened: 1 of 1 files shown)


control/control_plane.go
```diff
@@ -18,7 +18,6 @@ import (
 	"syscall"
 	"time"
```
mzz2017, line 20:
```suggestion

	"github.com/bits-and-blooms/bloom/v3"
```

cubercsl (approved):

PR #84, opened 2023-05-13
patch: search geodata at same dir with config first

mzz2017:
## Motivation

dae should search geodata in /etc/dae first.

## Summary

The same directory with config.dae will be given the highest priority for dae to search `geodata`.


yqlbu (approved):
LGTM.

PR #87, opened 2023-05-14
fix(dns): cache would never expire caused by #63 by accident

### LINKED ISSUE #83: [Support Request] DNS Cache problem
## What happened:
Does dae have DNS cache? I have DDNS configured on my router and nslookup can return the correct IP if not connected through dae. If connected via dae it always returns the old IP unless I restart the dae service
## Environment:

- Dae version (use `dae --version`):
- OS (e.g `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Others:

mzz2017:
### Background

Cache would never expire by accident.

### Checklist

### Full changelog

- Add expiration check when looking up the DNS cache.

### Issue reference

Fix #83
Related PR: https://github.com/daeuniverse/dae/pull/63/files#diff-7092e5e0724411e117b147fe12471c61569d904d911ef0f049885f9c1b4ce37eL135

### DIFF
--- a/control/dns_control.go
+++ b/control/dns_control.go
@@ -134,7 +134,7 @@ func (c *DnsController) LookupDnsRespCache(qname string, qtype dnsmessage.Type) 
 	c.dnsCacheMu.Lock()
 	cache, ok := c.dnsCache[c.cacheKey(qname, qtype)]
 	c.dnsCacheMu.Unlock()
-	// We should make sure the remaining TTL is greater than 120s (minFirefoxCacheTimeout), or
+	// We should make sure the cache does not expire, or
 	// return nil and request a new lookup to refresh the cache.
 	if ok && cache.Deadline.After(time.Now()) {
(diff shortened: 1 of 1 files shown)


control/dns_control.go
```diff
@@ -134,9 +134,9 @@ func (c *DnsController) LookupDnsRespCache(qname string, qtype dnsmessage.Type)
 	c.dnsCacheMu.Lock()
 	cache, ok := c.dnsCache[c.cacheKey(qname, qtype)]
 	c.dnsCacheMu.Unlock()
-	// We should make sure the remaining TTL is greater than 120s (minFirefoxCacheTimeout), or
+	// We should make sure the cache does not expire, or
```
mzz2017, line 137:
```suggestion
	// We should make sure the cache did not expire, or
```

yqlbu (approved):

PR #94, opened 2023-05-18
feat: add uTLS support

AkinoKaede:
### Background

For bypass the TLS Fingerprint DPI, I add uTLS support for `dae`.

### Checklist


component/outbound/dialer/dialer.go
```diff
@@ -39,6 +39,12 @@ type GlobalOption struct {
 	CheckTolerance    time.Duration
 	CheckDnsTcp       bool
 	AllowInsecure     bool
+	TlsImplementation string
+	Utls              Utls
```
mzz2017, line 43: Can we remove this to reduce parameter number and give it a default value?

AkinoKaede: `TlsImplementation·`should be held so that we can easily add other TLS implementations. But I can change `Utls.Imitate` to `UtlsImitate`. I think it should be completely up to the user to choose the `Client Hello ID` so that the user knows what they are doing.

mzz2017: Great. I think UtlsImitate is better! Thanks ~

mzz2017: By the way, could you please add some annotations(and all option values) in example.dae and also in config/desc.go?

AkinoKaede: Of course.

mzz2017:
@AkinoKaede Yes. Add a markdown please.

mzz2017:
@yqlbu I have no tls node. Would you like to test this PR? Thank you!

mzz2017 (approved):

PR #119, opened 2023-06-04
feat: latency-based failover

### LINKED ISSUE #118: Request failover feature
## What feature you would like us to integrate into the dae project

I want to treat my US node (with higher average latency) as main use, and if it is down, failover to HK node (with lower average latency).

## Why is this needed

My US node has more traffic and bandwidth.

mzz2017:
### Background

I want to treat my US node as main use, and if it is down, failover to HK node. **In this way, US nodes with high latency can be used preferentially.**

### Idea

Normalize all failover situations to latency-based solution.

This solution has following advantages:

1. If the US node is under DDOS, the connectivity is not stable and the node will be up and down. In this situation, a timeout (10000ms) will be added to the latency every down and this node will have lower priority. The priority will recover automatically if DDOS is over.
1. Reuse current policies: min, min_avg10, min_moving_avg, ...

```nginx
group {
  my_group {
    filter: name(my_HKnode) [add_latency: 300ms]
    filter: name(my_USnode, node_c) [add_latency: -500ms]
    filter: name(node_d, node_e, node_f)
    policy: min_avg10
  }
}
```

Thus, we need to introduce the annotation grammar(`[add_latency: 50ms]`).

### Checklist

- [x] The Pull Request has been fully teste

### Full changelog

- Support `annotation` syntax for `declaration` (`[key: value, ...]`).
- Support multiple filters in one group.
- Support `add_latency` annotation for `filter` and apply it to corresponding dialers (if multiple filter hits, only the first works).

### Issue reference

Fix #118


mzz2017:
Some changes need to do. The added latency should be in group level.

mzz2017:
One more thing:
It changed behavior that "no filter" means "no filter" to "no filter" means "no node". It should be fixed.

mzz2017:
"Group level 'add_latency'" and "no filter problem" have been fixed. Ready for review.

mzz2017:
image

mzz2017:
Another test result:
image

PR #80, opened 2023-05-13
feat: support iptables tproxy

mzz2017:
## Motivation

Support to use iptables to redirect traffic to dae, which allows user to use another method to redirect traffic to dae.

## Modification

1. Give an option to allow user to disable tproxy port protection. Tproxy port protection strictly checks all packets to tproxy port of dae.
2. Allow to set so_mark from dae to avoid traffic loop.

## Example

To redirect all traffic to dae:

```bash
global {
    tproxy_port: 12345
    tproxy_port_protect: false
    so_mark_from_dae: 0x80
    #...
}
```

```bash
ip rule add fwmark 0x40/0xc0 table 100
ip route add local 0.0.0.0/0 dev lo table 100

iptables -t mangle -N TP_PRE
iptables -t mangle -A PREROUTING -j TP_PRE
iptables -t mangle -N TP_OUT
iptables -t mangle -A OUTPUT -j TP_OUT
iptables -t mangle -N TP_MARK
iptables -t mangle -A TP_MARK -p tcp -m tcp --tcp-flags FIN,SYN,RST,ACK SYN -j MARK --set-xmark 0x40/0x40
iptables -t mangle -A TP_MARK -p udp -m conntrack --ctstate NEW -j MARK --set-xmark 0x40/0x40
iptables -t mangle -A TP_MARK -j CONNMARK --save-mark --nfmask 0xffffffff --ctmask 0xffffffff
iptables -t mangle -A TP_OUT -m mark --mark 0x80/0x80 -j RETURN
iptables -t mangle -N TP_RULE
iptables -t mangle -A TP_OUT -p tcp -m addrtype --src-type LOCAL ! --dst-type LOCAL -j TP_RULE
iptables -t mangle -A TP_OUT -p udp -m addrtype --src-type LOCAL ! --dst-type LOCAL -j TP_RULE
iptables -t mangle -A TP_PRE -i lo -m mark ! --mark 0x40/0xc0 -j RETURN
iptables -t mangle -A TP_PRE -p tcp -m addrtype ! --src-type LOCAL ! --dst-type LOCAL -j TP_RULE
iptables -t mangle -A TP_PRE -p udp -m addrtype ! --src-type LOCAL ! --dst-type LOCAL -j TP_RULE
iptables -t mangle -A TP_PRE -p tcp -m mark --mark 0x40/0xc0 -j TPROXY --on-port 12345 --on-ip 127.0.0.1
[...]

PR #155, opened 2023-06-25
fix: bad connectivity check of dns(udp) caused by #80

mzz2017:
### Background

#80 caused failed connectivity check, which impacts node protocols with bad UDP support like VLESS(XUDP).

In detail, it causes checking only UDP DNS connectivity.

Affected range:

1. v0.2.0rc1
2. VLESS protocol

### Checklist

### Full changelogs

- Enable correct TCP DNS check.


yqlbu:
Test result from the stakeholder:

image

yqlbu (approved):

Issue #156, opened 2023-06-26
CI Roadmap


yqlbu:
## Context

Constantly improve CI quality build.

## TODOs

- [x] Separate `release` build from regular build
- [x] Separate `pr-based` build from regular build
- [x] Add automation of `example.dae` to `generate-changelogs` workflow
- [x] Add `daily-preview` build
- [x] Use `shared` workflows/actions to chain reusable steps to various downstream workflows ==> `CI Modularity`

yqlbu:
According to @MarksonHon, quoted:

> 构建与压缩这两个步骤可以用脚本去操作，方便复用

yqlbu:
Task - `Add` automation of example.dae to generate-changelogs workflow` has been implemented in https://github.com/daeuniverse/changelogs-generator-action/pull/10

yqlbu:
Task - `Add daily-preview build` has been implemented in #189

yqlbu:
Task - Use `shared` workflows/actions to chain reusable steps to various downstream workflows ==> `CI Modularity` has been implemented in #190

yqlbu:
As all TODOs have been implemented, close this issue.

PR #159, opened 2023-06-27
ci: add generate-changelogs workflow

yqlbu:
### Background

Add generate-changelogs workflow; it is identical to the one we have for `daed`; 

> **Note**: Only when we design and implement `CI modularity`, will we use a single `unified` workflow to generate changelogs for future releases.

### Checklist

- [x] There's an entry in the CHANGELOG

### Full changelogs

- ci: add generate-changelogs workflow
- chore(changelogs): add changelogs placeholder

### Issue reference

As part of #156


kunish (approved):
We've already tested and verified this on daed, since the action is identical to the one daed is using, so it should be safe to assume this would work.

PR #164, opened 2023-06-29
fix: should allow fallbacking ip version if dialing domain

### LINKED ISSUE #152: Happyeyeballs is not fully supported by some softwares
## What would you like us to improve

If local has ipv6 addresses and routes and remote proxy doesn't, eBPF program will disable the ipv6 traffic to this remote proxy.

However, some softwares have bad happy eyeballs support and always retry.

### Node.js

Occurs since v17.
Fixed in v20.

https://github.com/nodejs/node/issues/41625
https://github.com/nodejs/node/pull/44731

### Powershell

Occurs on but not only v7.3.4.

image

### wget

No version context was given.

Wait 60s for timeout and fallback.

image

## Why is this needed

We should disable dropping ipv6 traffic if node has no ipv6 connectivity (at least) in domain++ mode.

mzz2017:
### Background

If the local host has IPv6 addresses but the remote proxy does not, `dae` drops all the IPv6 traffic onto the remote proxy. This behavior impacts lots of software that does not support happy eyeballs.

In fact, if we send the domain to the remote proxy, it will re-select ip version and this problem will be resolved.

In the past, `dae` did not support domain sniffing. The developer hope to solve this problem by happy eyeballs, but now it seems that this change needs to be optimized, and only when dialing ip instead, maintain the previous behavior.

### Checklist

### Full changelogs

- If `dial_mode` is not `ip`, dae removes notifying the kernel program to block the unreachable triples (group, ip version, l4 protocol).
- If the `dial` target is not an ip, dae trys to fallback to another ip version to dial.

### Issue reference

Fix #152

### DIFF
--- a/common/consts/ebpf.go
+++ b/common/consts/ebpf.go
@@ -131,6 +131,16 @@ const (
 	IpVersion_X IpVersionType = 3
 )
 
+func (v IpVersionType) ToIpVersionStr() IpVersionStr {
+	switch v {
+	case IpVersion_4:
+		return IpVersionStr_4
+	case IpVersion_6:
+		return IpVersionStr_6
+	}
+	panic("unsupported ipversion")
+}
+
 var (
 	BasicFeatureVersion = internal.Version{5, 2, 0}
 	// Deprecated: Ftrace does not support arm64 yet (Linux 6.2).
--- a/component/outbound/dialer_group.go
+++ b/component/outbound/dialer_group.go
@@ -6,6 +6,7 @@
 package outbound
 
 import (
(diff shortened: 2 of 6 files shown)


mzz2017:
图片

MarksonHon:
图片
PowerShell also works well now

MarksonHon:
图片
Poor wget works well now

MarksonHon (commented):
tested, it works.

mzz2017:
@MarksonHon  LGTM. Thanks bro.

PR #166, opened 2023-06-30
feat: support to bind to lan lazily and re-bind to lan

mzz2017:
### Background

Sometimes, we want to bind a `LAN` interface once it is brought up (NewLink).

For example, I want to bind dae to `ap0` once I start to share my hotspot. However, I do not exactly know when in the future I will start to share my hotspot and create an `ap0`. dae will fail to start if fail to bind any LAN interfaces.

图片

Another problem is that if my `ap0` is deleted, when `ap0` is created in the future, `dae` will not bind to it again.

This PR fixes it.

### Checklist

### Full changelogs

- If fail to bind due to `link not found`, wait for `RTM_NEWLINK` and bind to it.
- If a link is deleted, wait for `RTM_NEWLINK` and bind to it.

### DIFF
--- a/control/control_plane.go
+++ b/control/control_plane.go
@@ -203,11 +203,7 @@ func NewControlPlane(
 		}
 		global.LanInterface = common.Deduplicate(global.LanInterface)
 		for _, ifname := range global.LanInterface {
-			if global.AutoConfigKernelParameter {
-				SetSendRedirects(ifname, "0")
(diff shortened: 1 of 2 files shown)


mzz2017:
This was tested by CTO.

yqlbu:
@mzz2017 Cool, let’s merge it.

PR #203, opened 2023-07-10
fix: truncated dns resp

mzz2017:
### Background

image

If a DNS response is too big (greater than 512 Bytes), it would be truncated. So the `unpacking` would fail.

This PR enlarges the buffer size to 2048. Hope it is not expensive.

### Checklist

### Full changelogs

- Enlarge DNS response buffer to `2048`.
- Upgrade `softwind` to support `pool.GetFullCap`.

### DIFF
--- a/common/consts/net.go
+++ b/common/consts/net.go
@@ -1,0 +1,5 @@
+package consts
+
+const (
+	EthernetMtu = 1500
+)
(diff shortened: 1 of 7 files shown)


mzz2017:
Test result:
image

PR #204, opened 2023-07-10
fix/optimize/refactor(udp): fix potential stuck UDP and optimize reroute logic

mzz2017:
#### Fix

The stuck UDP dialing would block all other UDP dialing. This causes UDP to fail to work. This PR fixes it.

Range:

1. This problem impacts all previous versions.
2. This problem only impacts UDP except DNS.

#### Optimize

`domain++` may reroute every UDP packet, which may consume many resources and result in unexpected results.

This PR optimizes it. If a UDP connection has already been established, we shouldn't reroute it (find another outbound/dialer) while we should reuse the dialer to write a request.

### Changelogs

- Force to give 8s timeout for TCP/UDP dialers.
- Use fine-grained lock for UDP.
- Only re-route for the new UDP connection.

### Checklist

### DIFF
--- a/common/consts/dialer.go
+++ b/common/consts/dialer.go
@@ -7,6 +7,7 @@ package consts
 
 import (
 	"net/netip"
+	"time"
 
 	"golang.org/x/sys/unix"
 )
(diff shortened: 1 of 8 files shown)


yqlbu (approved):
This PR has been tested, working fine in my environment.

PR #207, opened 2023-07-11
fix: No AddrPort presented

mzz2017:
### Background

image

image

This problem deepened since #203. However, this problem exists for a long time.

This PR fixes it.

### Checklist

### Full changelogs

- Pass in a copy of `oob` into the goroutine instead of reusing it.


mzz2017:
Test result:
image

PR #221, opened 2023-07-16
fix: support to bind dae to wg lan (and other tun)

mzz2017:
### Background

When using the dae's bind-lan functionality, the binding fails to work properly when attempting to bind a WireGuard interface. As a result, the traffic from the WireGuard interface is not being routed through the dae proxy.

See more #191.

### Checklist

- [x] The Pull Request has been fully tested
- [ ] There's an entry in the CHANGELOGS
- [ ] There is a user-facing docs PR against https://github.com/daeuniverse/dae

### Full changelogs

- Support to check link type before binding to it.
- Support link type `none`.
- Refine `ebpf` code.

### Issue reference

Closes #191.

PR #225, opened 2023-07-16
fix(ebpf): udp problem caused by #221 by accident

### LINKED ISSUE #224: Broken UDP caused by #221
## What happened

```
WARN No AddrPort presented: reading map: key [[::ffff:192.168.31.194]:60563, 17, 198.211.120.59:3478]: lookup: key does not exist, data is too short to parse AddrHdr
```

## How to reproduce it (as minimally and precisely as possible):

`stun` or `dns lookup`

## Environment

- Dae version (use `dae --version`):
- OS (e.g `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Others:

mzz2017:
### Background

In #221, I incorrectly parse UDP header as TCP header, which causes a failure in parsing five tuples and results in bad connectivity of UDP. 

### Checklist

### Full changelogs

- Correct UDP parsing.

### Issue reference

Closes #224


mzz2017:
Test result:
图片

PR #251, opened 2023-08-01
optimize: force to choose one if there is only one node in the group

mzz2017:
### Background

Should fallback to `fixed(0)` if there is only one node in this group and even if it not alive.

### Checklist

### Full changelogs

- Change the logic in node `select`.

### Issue reference

N/A

### Test Result

N/A


mzz2017:
Test Result:
image

yqlbu (approved):

PR #257, opened 2023-08-05
fix: bad ipversion selection caused by #251

mzz2017:
### Background

dae always chooses ipv6 DNS to dial which causes `no network`.

Reproduce conditions:

1. The group only has one node.
1. DNS upstream uses domain, thus dae can select between ipv4 and ipv6.
1. Remote proxy doesn't support ipv6.

### Checklist

### Full changelogs

- Return 10*second (timeout) instead of 0ms as latency if there is no alive dialer for a given network type.
- Refine code.
- Upgrade go module softwind.

https://github.com/daeuniverse/dae/pull/257/files#diff-60d7f789fcd6cccaaeca0f229f53ffdff0ce5586805cb6f1908db25619f361f8R236-R245

### Related PR

#251


yqlbu (approved):
Looks good

kunish (approved):

PR #280, opened 2023-08-14
Makefile: remove build with pie mode

MarksonHon:
### Background

If build with pie, cgo will never be disabled, then OpenWrt, Alpine Linux and other systems based on MUSL C will not load dae/dae-wing.

### Checklist

### Full Changelogs

- [Implement ...]

### Issue Reference

Closes #_[issue number]_


yqlbu:
Duplicated with #282

PR #288, opened 2023-08-18
feat: dae can call now

troubadour-hell:
### Background

As a dae, it should call.

### Checklist

### Full Changelogs

- [Implement ...]

### Issue Reference

Closes #_[issue number]_

### DIFF
--- a/cmd/call.go
+++ b/cmd/call.go
@@ -13,8 +13,8 @@
  )
  
  var (
-	 callCmd = &cobra.Command{
-		 Use:   "call",
+	 honkCmd = &cobra.Command{
+		 Use:   "honk",
 		 Short: "Let dae call for you.",
 		 Run: func(cmd *cobra.Command, args []string) {
 				 fmt.Println("Honk! Honk! Honk! This is dae!")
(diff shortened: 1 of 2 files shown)


mzz2017:
@troubadour-hell Use 'honk' instead?

troubadour-hell:
> 

Done.

troubadour-hell:
> @troubadour-hell Use 'honk' instead?

done.

cmd/call.go
```diff
@@ -0,0 +1,29 @@
...
+	 "github.com/spf13/cobra"
+ )
+ 
+ var (
+	 honkCmd = &cobra.Command{
+		 Use:   "honk",
+		 Short: "Let dae call for you.",
+		 Run: func(cmd *cobra.Command, args []string) {
+				 fmt.Println("Honk! Honk! Honk! This is dae!")
+				 os.Exit(1)
```
mzz2017, line 21: 1 indicates errors. We should use os.Exit(0) here.

BTW, do you have tested it? Please paste the result. Thank you.

mzz2017 (changes requested):

cmd/call.go
troubadour-hell: With exit(1), it works like:
image
And it was changed to exit(0) in this fix commit.

mzz2017:
Interesting PR. Thank you!

mzz2017 (approved):

PR #292, opened 2023-08-19
feat: support ToS routing rule

mzz2017:
### Background

Support `tos` routing rule.

According to <http://libtorrent.org/reference-Settings.html#peer_tos>, the default value of `dscp` and `tos` is 0x04 in libtorrent.

This PR gives a BitTorrent bypass method.

Example:

```
routing {
  tos(4) -> direct
}
```

### Checklist

- [x] The Pull Request has been fully tested
- [ ] There's an entry in the CHANGELOGS
- [x] There is a user-facing docs PR against https://github.com/daeuniverse/dae

### Full Changelogs

- [Implement ...]

### Issue Reference

Closes #_[issue number]_

PR #294, opened 2023-08-20
refactor/fix: match dscp instead of tos

mzz2017:
### Background

A problem was introduced in #292. This is because we added `dscp` as a member of `struct tuples`. However, `struct tuples` is used for five tuples matching. When `dscp` is added, packets with the non-zero dscp will be rejected due to matching failures. This problem impacts packets unmatched in eBPF programs.

Another change is that, we should match dscp instead of tos.

```
dscp = tos >> 2
```

### Checklist

### Full Changelogs

- [Implement ...]

### Issue Reference

Closes #_[issue number]_

### Test Result

image


mzz2017:
Test Result:
image
image

mzz2017:
Test Result 2 (tcp, ipv6):
image

yqlbu (approved):
Cool feature!

PR #329, opened 2023-09-11
docs: refine DNS example

EkkoG:
### Background

Since `qname` condition is more slow than `ip` condition, so the DNS example should move the `ip` condition to the start, for most DNS request, it will only need check if IP is private.

### Checklist

### Full Changelogs

- Refine DNS example, better performance

### Issue Reference

no

### Test Result

no

### DIFF
--- a/docs/en/README.md
+++ b/docs/en/README.md
@@ -180,7 +180,7 @@ dns {
     }
     response {
       upstream(googledns) -> accept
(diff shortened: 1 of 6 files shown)


mzz2017 (approved):
LGTM.

mzz2017:
Thanks for your insightful catch.

Issue #368, opened 2023-12-18
[Enhancement] Add CI test


jschwinger233:
### Improvement Suggestion

My understanding is dae is supposed to run on Linux >= 5.8, let's (let me) add some test via github action against different kernels.

I am proposing to use cilium/little-vm-helper action for running VM inside a github action container.

If maintainers are in favor of this plan, I can draft a PR for review in days.

daebot:
❣️ This issue is marked as `wontfix` as you have not yet starred this repo. Please kindly consider giving a star to this repo. Your support means a lot to us. Thanks for your understanding. After you become a stargazer, please also reply to this message with the keyword `understood`. Afterward, I will reopen this issue for you. Once again, your support is much appreciated. Cheers.

jschwinger233:
> ❣️ This issue is marked as `wontfix` as you have not yet starred this repo. Please kindly consider giving a star to this repo. Your support means a lot to us. Thanks for your understanding. After you become a stargazer, please also reply to this message with the keyword `understood`. Afterward, I will reopen this issue for you. Once again, your support is much appreciated. Cheers.

understood

daebot:
✅ Thanks for giving a star to the repository. Your support means a lot to us. I removed the `wonfix` mark for you. Hopefully, someone can dig in and resolve the issue for you.

yqlbu:
@jschwinger233 cc @mzz2017 I think it is a nice to have feature.

mzz2017:
LGTM. I'm glad to hear this exciting work.

Draky50110:
@daebot if you want, I can freely star your repo, as long as you either pay me or suck my dick.

Hope your project will die.

Don't thanks me, as begging a star is a pleasure to see in this shittification of the world.

nukeop:
Reported for farming.

Draky50110:
@nukeop Thanks, I will star your project too.

nukeop:
Since the offending comments were redacted, I'm gonna post what they looked like before.

<img width="391" alt="image" src="https://github.com/daeuniverse/dae/assets/12746779/c647314a-8d69-4dd0-a689-55e178fdd287">

baglayan:
I am speechless.

shafik:
> Reported for farming.

How do you report for farming?

Starpelly:
People are so rude. Don't they know open source developers need those juicy stars to eat?

wnm210:
yum

[...]

PR #373, opened 2023-12-19
ci: Test against various kernels on PR

jschwinger233:
Closes: #368


jschwinger233:
New workflow requires approval for running. I ran them on my personal fork https://github.com/jschwinger233/dae/pull/1, it looks code logic is fine but `apt upgrade` broke from time to time on 5.15/5.10

jschwinger233:
Alright, looks better: https://github.com/jschwinger233/dae/actions/runs/7262360947/job/19785369670?pr=1

jschwinger233:
> are these proposed changes mainly for testing purpose?

That's right. Although not all cases are covered (only TCP + dip filter + socks5 are tested), this workflow gives a good sense of whether dae works under different kernels. If bpf verifier complains due to complexity issues, this workflow definitely will catch them. We can also extend test suite as a follow-up if we want.

.github/workflows/test.yml
```diff
@@ -0,0 +1,160 @@
+name: Build and Test
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
```
yqlbu, line 7:
```suggestion
    paths:
      - "**/*.go"
      - "**/*.c"
      - "**/*.h"
      - "go.mod"
      - "go.sum"
      - ".github/workflows/test.yml"
```

Add trigger conditions (based on files).

.github/workflows/test.yml
```diff
@@ -0,0 +1,160 @@
+name: Build and Test
+
+on:
+  push:
+    branches: [ main ]
```
yqlbu, line 5:
```suggestion
```

For trouble shooting purpose, we may do the kernel-related testing in the PR level to ensure we fix the incompatible issue before we merge it to main branch.

yqlbu (changes requested):

yqlbu (commented):
@jschwinger233 Thanks for the fabulous work. One little comment/suggestion from my end, would you like to rename `test.yml` to `kernel-test.yml` to clarify the purpose of the test.

yqlbu (approved):
LGTM

yqlbu:
Once again appreciate all the efforts.

Issue #384, opened 2023-12-28
[Bug Report] Pid matching misfunctions when running dae inside a pid-namespaced container


jschwinger233:
### Checks

- [X] I have searched the existing issues
- [X] I have read the documentation
- [X] Is it your first time sumbitting an issue

### Current Behavior

Running dae inside a container with an independent pid namespace leads to infinite loop. This is because cgroup/* bpf programs record host pid while dae setup namespaced pid by `os.GetPid()`.

Likely it's not a bug, as I am not aware of any solution except:
1. we match process name instead of pid
2. we ditch the pid matching to be in favor of mark matching (to distinguish dae traffic)

Still, the general pid routing fails to work properly when inside a pid-namespaced container.

Not saying we must take care of this, just open an issue for record and awareness.

### Steps to Reproduce

1. Prepare the conf:

```
global {
    tproxy_port: 12345
    log_level: trace
    wan_interface: auto
    allow_insecure: false
}

node {
    local: 'socks5://socks5:1080'
}

group {
    proxy {
        policy: min_moving_avg
    }
}

routing {
    fallback: proxy
}
```

2. Run dae inside a pid-namespaced container

```
# assume we have "dae" binary under pwd
docker run -td --name dae --privileged -v $(pwd):/host ubuntu:22.04 bash
docker exec dae mount bpffs -t bpf /sys/fs/bpf
docker exec dae mount -t debugfs none /sys/kernel/debug/
docker exec -it docker exec -it dae /host/dae run -c /host/conf.dae
```

3. dae gets stuck into infinite loop

### Environment

- **Dae version (use `dae --version`)**: dae version unstable-20231221.r627.84c8eeb
- **OS (e.g `cat /etc/os-release`)**: Ubuntu 22.04.3 LTS
- **Kernel (e.g. `uname -a`)**: Linux gray-Latitude-5530 6.2.0-39-generic #40~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Nov 16 10:53:04 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
- **Others**: None

mzz2017:
![image](https://github.com/daeuniverse/dae/assets/30586220/28b5eb0a-b76f-4f36-9ccf-81b6549f4baf)

https://hub.docker.com/r/daeuniverse/dae

Yes. It is what we recommend.

jschwinger233:
@sumire88 Not really, this issue if for record and awarenes in case other users hit it.

[...]

PR #386, opened 2023-12-29
ci: Add more kernel tests

jschwinger233:
Covers IPv4 + IPv6 for TCP + UDP under WAN + LAN.

IPv6 UDP is buggy so I skipped that: #102 #387

### DIFF
--- a/.github/workflows/kernel-test.yml
+++ b/.github/workflows/kernel-test.yml
@@ -214,3 +214,121 @@ jobs:
               echo "Ingore failure until https://github.com/daeuniverse/dae/issues/387 is resolved"
               true
             fi
+
+      - name: Setup LAN
+        uses: cilium/little-vm-helper@908ab1ff8a596a03cd5221a1f8602dc44c3f906d # v0.0.12
+        with:
+          provision: 'false'
+          cmd: |
+            set -ex
+            cd /host
+
+            docker restart -t0 dae
+            docker exec dae apt install -y iproute2 iptables iputils-ping
+
+            cat >lan.bash <<!
+            set -ex
+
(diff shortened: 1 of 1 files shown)


.github/workflows/kernel-test.yml
```diff
@@ -135,30 +157,111 @@ jobs:
             }
 
             routing {
-                dip(1.1.1.1) -> proxy
-                fallback: direct
+                pname(dae) -> direct
```
mzz2017, line 160: Why we need this?

jschwinger233: https://github.com/daeuniverse/dae/issues/384

Pid matching doesnt work inside container

jschwinger233:
IPv6 keeps failing, in my opinion this is because github action doesn't IPv6 outbound to world: https://github.com/actions/runner-images/issues/668.

I therefore plan to test IPv6 3-way handshake only, to make sure packets originated by a local process can finally arrive at dae process.

As for IPv6 UDP, it seems to be buggy on main. Will troubleshot that later.

jschwinger233:
好耶，IPv6 TCP for WAN / LAN 也测过了，ready for review.

mzz2017:
@jschwinger233 Exciting news. What about ipv6 udp?

mzz2017:
@jschwinger233 I think we should also test it whatever it has bug or not.

jschwinger233:
@mzz2017 IPv6 UDP testcase added.

mzz2017:
Cool. Thanks for your brilliant work. I think it is ready to merge.

mzz2017 (dismissed):

PR #399, opened 2024-01-01
fix: ip rule is not added when binding wan

mzz2017:
### Background

Since #383 added the mark-based tproxy support on wan interfaces, we should also add ip rule when binding WAN interfaces.

### Checklist

### Full Changelogs

- [Implement ...]

### Issue Reference

Closes #_[issue number]_


jschwinger233 (approved):
Thanks!

PR #401, opened 2024-01-02
fix(bpf): sk_lookup_udp for listener only

jschwinger233:
### Background

Consider two consecutive `dig @1.1.1.1 one.one.one.one`: dae will create a UDP socket binding `1.1.1.1:53` for the 1st DNS query, then by the time the 2nd one arrives at wan_egress@eth0, sk_lookup_udp will match that newly created UDP socket by mistake. 

As lookup for established UDP is not necessary at all, this PR skips that to avoid this tragedy bug.

Fixes: #383 

### Checklist

### Full Changelogs

- [Implement ...]

### Issue Reference

Closes #_[issue number]_

### DIFF
--- a/control/kern/tproxy.c
+++ b/control/kern/tproxy.c
@@ -1870,24 +1870,29 @@ int tproxy_wan_ingress(struct __sk_buff *skb) {
 
   skb->mark = TPROXY_MARK;
   struct bpf_sock_tuple tuple = {};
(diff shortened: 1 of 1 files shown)


mzz2017:
Great, it's so smooth now.

mzz2017 (approved):
Nice catch!

mzz2017:
> Has this patch been fully tested yet?

Yes. Let's move on.

PR #403, opened 2024-01-02
docs(troubleshooting.md): add firewalld related docs

mzz2017:
### Background

由于新版本（v0.5.0+）不再使用两次 NAT 方案，所以放行 port 12345 的方案无法使得防火墙放行 dae 流量。

firewalld 会拦截 reroute packets，firewalld 无法支持匹配 mark，因此只能使用 nft 来加入放行规则。

The new version (v0.5.0+) no longer uses the double NAT scheme, so the solution to allow port 12345 cannot make the firewall allow dae traffic.

Firewalld will block reroute packets. Firewalld does not support matching marks, so only NFT (nftables) can be used to add allow rules.

### Checklist

### Full Changelogs

- [Implement ...]

### Issue Reference

Closes #_[issue number]_

### DIFF
--- a/docs/en/troubleshooting.md
+++ b/docs/en/troubleshooting.md
@@ -27,7 +27,7 @@ ufw
 firewalld
 ```
 
-**ufw**
+#### ufw
 
 UFW users may need some extra steps to make sure `Binding to LAN` working.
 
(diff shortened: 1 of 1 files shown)


mzz2017:
Thanks @jschwinger233

PR #418, opened 2024-01-08
fix/docs: fix the first check failure and possible dns leaking

mzz2017:
### Background

When local dns server is listening at :53 and resolv.conf points to it, the first connectivity check will fail. This is because when we use the local dns server to resolve the domain name of the node, the dae is not yet listening, but the ebpf program has been installed and dns request is redirected to dae, so the first dns request will fail.

If we add local dns servers to must_direct list, dns leaking possibly occurs.

### Checklist

### Full Changelogs

- Activate connectivity check after listening to avoid the first failure.
- Update docs and don't add these dns servers to must_direct list.

### Issue Reference

Closes #_[issue number]_

### Test Result

image

### DIFF
--- a/example.dae
+++ b/example.dae
@@ -143,20 +143,30 @@ dns {
         # According to the request of dns query, decide to use which DNS upstream.
         # Match rules from top to bottom.
         request {
+            # Lookup China mainland domains using alidns, otherwise googledns.
+            qname(geosite:cn) -> alidns
             # fallback is also called default.
-            fallback: alidns
-        }
-        # According to the response of dns query, decide to accept or re-lookup using another DNS upstream.
-        # Match rules from top to bottom.
-        response {
-            # Trusted upstream. Always accept its result.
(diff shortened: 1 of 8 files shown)


example.dae
```diff
@@ -143,20 +143,30 @@ dns {
...
+#        # Match rules from top to bottom.
+#        response {
+#            # Trusted upstream. Always accept its result.
+#            upstream(googledns) -> accept
+#            # Possibly polluted, re-lookup using googledns.
+#            ip(geoip:private) && !qname(geosite:cn) -> googledns
+#            # fallback is also called default.
+#            fallback: accept
+#        }
+#    }
```
yqlbu, line 169: Considered this is the `advanced` use cases. Shall we put them in https://github.com/daeuniverse/dae/blob/main/docs/en/configuration/routing.md?

We may create a separate block to include the proposed changes aboved.

mzz2017: I agree with you.

PR #446, opened 2024-01-27
fix: Create DaeNetns instance strictly once on reload

jschwinger233:
### Background

Previously DaeNetns instance can be re-created on reload, breaking the singleton pattern. This PR fixes it by introducing sync.Once.

### Checklist

### Full Changelogs

- [Implement ...]

### Issue Reference

Closes #_[issue number]_

### DIFF
--- a/control/netns_utils.go
+++ b/control/netns_utils.go
@@ -24,6 +24,7 @@ const (
 
 var (
 	daeNetns *DaeNetns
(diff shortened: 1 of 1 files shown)


QiuSimons:
it works, indeed.

mzz2017 (approved):

PR #505, opened 2024-04-23
fix: maintain udp conn direction to fix possible dns leaking

mzz2017:
### Background

In the past, the direction of the UDP connection was not maintained, so receiving a return packet would cause all subsequent packets of the same five-tuple to be sent directly, which is not in line with expectations. This PR will fix it.

### Checklist

### Full Changelogs

- [Implement ...]

### Issue Reference

Closes #_[issue number]_


wanlce (approved):
Test Passed!

PR #588, opened 2024-07-22
fix(cmd/run.go): always check whether network online

MarksonHon:
### Checklist

### Full Changelogs

- run.go: Network Manager's net-online check has a bug: you'll **be** online even if you have got no IP address, so we should always check whether network is online by dae itself.

### Issue Reference

Closes #_[issue number]_


mzz2017 (approved):
LGTM

PR #600, opened 2024-08-07
optimize(bpf): Alternative way to avoid parsing packet at dae0

### LINKED ISSUE #586: [Bug Report] v0.8.0rc1进程消失
### Checks

- [X] I have searched the existing issues
- [X] I have read the documentation
- [ ] Is it your first time sumbitting an issue

### Current Behavior

启动dae v0.8.0rc1后一段时间进程就会消失，而且ssh会短暂地无法连接，如果ssh已经连接就会显示`client_loop: send disconnect: Connection reset`。v0.7.0没问题

### Expected Behavior

正常运行

### Steps to Reproduce

1. 使用dae v0.8.0rc1
2. `dae run --config /etc/dae/config.dae`

### Environment

- **Dae version**: v0.8.0rc1
- **OS**: OpenWrt SNAPSHOT r26958-9d8d0509bf
- **Kernel**: Linux OpenWrt 6.6.40 #0 SMP Wed Jul 17 21:27:50 2024 x86_64 GNU/Linux

### Anything else?

_No response_

jschwinger233:
### Background

#562 caused more issues than benefits, such as #586. Unfortunately I failed to reproduce issues after several attempts. This PR reverts the changes, but introduced an alternative solution to avoid parsing packet at dae0_ingress.

~~The key idea is skbs at dae0_ingress are bound to be built from a local process (control plane process, dae), so we should be confident in using `skb->sk->{local_ip4,dst_ip4,local_ip6,dst_ip6}` to get tuples, instead of calling `parse_transport()` followed by `get_tuples()`.~~ This doesn't work because udp listener socket always has zero value as sk->dst_ip4. 

The key idea is to parse L3 by manually calling `bpf_skb_load_bytes` for twice only. The instruction number can be greatly reduced from 913 to 333 (jitted x64, kernel 6.5.0-1027-oem), -63% comparing to current main (using direct access). Before #562 the insn number can be as many as 1715, then it's -80%.

The subtle part is this PR deletes `l4proto` from redirect_tuple. This looks harmless as a TCP connection and a UDP "connection" with same src/dst ip should have **same source**, therefore they can safely share one redirect_entry.

One followup on me: review and fix ci kernel-test.

### Checklist

### Full Changelogs

- [Implement ...]

### Issue Reference

Closes #586

### DIFF
--- a/control/kern/tproxy.c
+++ b/control/kern/tproxy.c
@@ -126,7 +126,6 @@ union ip6 {
 struct redirect_tuple {
 	union ip6 sip;
 	union ip6 dip;
-	__u8 l4proto;
 };
 
 struct redirect_entry {
@@ -963,7 +962,6 @@ static __always_inline void prep_redirect_to_control_plane(
 		__builtin_memcpy(&redirect_tuple.dip, &tuples->five.dip,
 				 IPV6_BYTE_LENGTH);
 	}
-	redirect_tuple.l4proto = l4proto;
 	struct redirect_entry redirect_entry = {};
 
 	redirect_entry.ifindex = skb->ifindex;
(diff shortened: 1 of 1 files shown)


mzz2017 (approved):
LGTM!

mzz2017 (approved):

PR #628, opened 2024-09-13
optimize: replace c arithmetic operators with bitwise ones

Integral-Tech:
### Background

It is better to use bitwise operators instead of arithmetic ones when merging bit fields.

### Checklist

### Full Changelogs

- Replace C arithmetic operators with bitwise ones to improve performance


mzz2017 (approved):
LGTM.

PR #634, opened 2024-09-14
refactor: remove redundant assignments

Integral-Tech:
### Background

There are some redundant assignments in ```control/kern/tproxy.c```.
- Assigning the 2nd & 3rd enumeration constant by 2 and 3 is unnecessary. If an enumeration constant is not followed by ```= constant-expression```, its value is the value one greater than the value of the previous enumerator in the same enumeration.
- The variable ```ret``` is not used after being assigned.
- When ```to_read == MAX_ARG_SCANNER_BUFFER_SIZE```, the assignment is unnecessary.

### Checklist

### Full Changelogs

- Remove redundant assignments in in ```control/kern/tproxy.c```

### DIFF
--- a/control/kern/tproxy.c
+++ b/control/kern/tproxy.c
@@ -302,14 +302,14 @@ enum __attribute__((packed)) MatchType {
 
 enum L4ProtoType {
 	L4ProtoType_TCP = 1,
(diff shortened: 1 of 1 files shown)


mzz2017 (approved):
LGTM!

PR #694, opened 2024-11-07
refactor: remove redundant type casting

Integral-Tech:
### Checklist

### Full Changelogs

- Code refactoring: remove redundant type casting in C code, including but not limited to `void *`


trace/kern/trace.c
```diff
@@ -180,7 +180,7 @@ set_tuple(struct tuple *tpl, struct sk_buff *skb)
 		tpl->sport= BPF_CORE_READ(tcp, source);
 		tpl->dport= BPF_CORE_READ(tcp, dest);
 		bpf_probe_read_kernel(&tpl->tcp_flags, sizeof(tpl->tcp_flags),
-				    (void *)tcp + offsetof(struct tcphdr, ack_seq) + 5);
+				    tcp + offsetof(struct tcphdr, ack_seq) + 5);
```
jschwinger233, line 183: 这个不对吧。。。你运行一下试试呢

jschwinger233 (changes requested):

trace/kern/trace.c
Integral-Tech: I checked the documentation of GNU C:

https://www.gnu.org/software/c-intro-and-ref/manual/c-intro-and-ref.html#Void-Pointers

> In standard C, addition and subtraction are not allowed on void *, since the target type’s size is not defined in that case. Likewise, they are not allowed on pointers to function types. However, these operations work in GNU C, and the “size of the target type” is taken as 1 byte.

Integral-Tech: Sorry about that

PR #700, opened 2024-11-18
optimize: reduce memory usage by reading kallsyms only in the trace command

seiuneko:
### Background
Currently, reading kallsyms during program loading causes unnecessary high memory usage, while kallsyms is only used in the trace command. Therefore, this patch changes it to initialize kallsyms before executing the trace command.

### Checklist

### Full Changelogs

- read kallsyms only in the trace command

#### Configuration

```
global{ pprof_port: 12333 }
routing{}
```
Use `go tool pprof -http=:8080 http://localhost:12333/debug/pprof/heap` to generate the flame graph.

#### Screenshot

Lowers RSS usage from 194 MiB to 75 MiB.

Before:
pprof_before
htop_before

After:
pprof_after
htop_after

### DIFF
--- a/cmd/trace.go
+++ b/cmd/trace.go
@@ -31,6 +31,9 @@ func init() {
 	traceCmd := &cobra.Command{
 		Use:   "trace",
 		Short: "To trace traffic",
+		PreRun: func(cmd *cobra.Command, args []string) {
+			trace.ReadKallsyms()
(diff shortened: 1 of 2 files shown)


jschwinger233 (approved):

mzz2017:
Nice catch

PR #738, opened 2025-01-19
fix(config): correct the truncation of process name(comm name)

### LINKED ISSUE #736: [Bug Report] pname rules don't work with process name longer than 15 characters
### Checks

### Current Behavior

pname rules don't work with process name longer than 15 characters.
that means 
pname(systemd-resolved) or pname(systemd-networkd) doesn't work
we have to write for now:
pname(systemd-resolve) or pname(systemd-network)
which is counter-intuitive

I am developing a fix to it and will make a pr in a moment, this issue is for the pr to target.

### Expected Behavior

Any longer process names set by pname rules work. dae should truncate the name automatically.

### Steps to Reproduce

1. set pname(systemd-resolved or other process name longer than 15 characters) -> xxx
2. dae reload

### Environment

- **Dae version (use `dae --version`)**: 0.9.0
- **OS (e.g `cat /etc/os-release`)**: fedora 41
- **Kernel (e.g. `uname -a`)**: 6.12.9

### Anything else?

_No response_

xcpky:
Previously dae truncate the process names in pname configs to 16 bytes Now we truncate the name to 15 bytes because the length of the comm name of processes in linux is 15 bytes.

This makes
`pname(systemd-resolved)`
works the same as
`pname(systemd-resolve)`

Closes #736, #474

### Background

#736 

### Checklist

### Full Changelogs

- truncate the process name to 15 characters(previously 16)
- issue the truncation info log whenever the process name is longer than 15 characters(previously 16)

### Issue Reference

#736 
#474 

Closes #_[736]_

### Test Result

add pname(systemd-resolved) -> must_direct
then all the dns traffic through systemd-resolved goes out directly
but before the PR, it doesn't.

### DIFF
--- a/component/routing/function_parser.go
+++ b/component/routing/function_parser.go
@@ -112,8 +112,8 @@ func ProcessNameParserFactory(callback func(f *config_parser.Function, procNames
 	return func(log *logrus.Logger, f *config_parser.Function, key string, paramValueGroup []string, overrideOutbound *Outbound) (err error) {
 		var procNames [][consts.TaskCommLen]byte
 		for _, v := range paramValueGroup {
-			if len([]byte(v)) > consts.TaskCommLen {
-				log.Infof(`pname routing: trim "%v" to "%v" because it is too long.`, v, string([]byte(v)[:consts.TaskCommLen]))
+			if len([]byte(v)) > consts.TaskCommLen - 1 {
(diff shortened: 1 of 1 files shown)


mzz2017 (approved):
LGTM

mzz2017:
Thanks for your contribution!

PR #908, opened 2025-12-21
fix: support absolute paths in include directive

### LINKED ISSUE #906: [Bug Report] `include` directive silently fails when using absolute paths
### Checks

### Current Behavior

When using an absolute path in the `include` directive, the included file is silently ignored without any error or warning message.

For example, with this configuration:

```
include {
  '/etc/dae/sub.dae'
}
```

The file `sub.dae` is **not loaded**, but dae produces **no error** of it, but errors about missing definitions in the included file.

### Expected Behavior

Abosulte paths in `include` should work correctly, or at least show a clear warning in docs like "absolute paths not supported in include directive".

I expected the fix because I am configuring my NixOS and encrypting subscription information, which is not appropriate to use relative path for the main config. (Probably I am doing in the wrong way.)

### Steps to Reproduce

1. Create a main config file `config.dae`:
   ```
   include {
       '/etc/dae/sub.dae'
   }

   global {
       tproxy_port: 12345
   }
   ```
2. Create a sub config file `sub.dae`:
   ```
   group {
       mygroup {
           # some proxies
       }
   }
   ```
3. Set proper permissions and then validate or run.
4. Observe that `mygroup` is not loaded.

### Environment

- **Dae version (use `dae --version`)**: unstable-20251103.r826.7e67e31
- **OS (e.g `cat /etc/os-release`)**: Ubuntu 24.04.3 LTS
- **Kernel (e.g. `uname -a`)**: 6.6.87.2-microsoft-standard-WSL2
- **Others**: NA

Note that the problem is found during my NisOS configuration, so probably it's not a WSL problem.

### Anything else?

In `config_merger.go` the `dfsMerge` function processes include paths like this:

```go
case *config_parser.Param:
    nextEntry := v.String(true, false)
    patterEntries = append(patterEntries, filepath.Join(m.entryDir, nextEntry))
```

for an absolute path `/etc/dae/sub.dae`, it will just result in `/etc/dae/etc/dae/sub.dae`. And after that, the `filepath.Glob(pattern)` will not throw any errors or warnings. So just consider check if the path is absolute before joining.

tsx8:
Fixes #906

Check if the include path is absolute before joining it with the entry directory.

### DIFF
--- a/docs/en/configuration/separate-config.md
+++ b/docs/en/configuration/separate-config.md
@@ -21,12 +21,22 @@ Directory Structure:
 
 Config files:
 
+Notes about `include` paths:
+
+- Relative paths (for example `config.d/*.dae`) are resolved relative to the directory of the *entry* config file (the file passed to `dae -c ...`), not relative to the current working directory.
+- Absolute paths (for example `/etc/dae/config.d/*.dae`) are used as-is.
+- For security reasons, dae only allows included files under the entry config directory.
+
 ```jsonc
 # config.dae
 
 # load all dae files placed in ./config.d/
 include {
+    # Relative path example:
     config.d/*.dae
+
+    # Absolute path example:
+    /etc/dae/config.d/*.dae
 }
 global {
     tproxy_port: 12345
--- a/example.dae
+++ b/example.dae
@@ -1,3 +1,20 @@
+#
+# You can split the configuration into multiple files with `include { ... }`.
+#
+# Notes about include paths:
+# - Relative paths (e.g. `config.d/*.dae`) are resolved relative to the directory of the entry config file
+#   (the file you pass to `dae -c ...`), NOT relative to your current working directory.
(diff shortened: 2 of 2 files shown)


MarksonHon:
Please add an example of absolute path in `example.dae`, and you should write some notes to make  the difference of absolute path and relative path clearly:

https://github.com/daeuniverse/dae/blob/main/example.dae#L136

tsx8:
> Please add an example of absolute path in `example.dae`, and you should write some notes to make the difference of absolute path and relative path clearly:
> 
> https://github.com/daeuniverse/dae/blob/main/example.dae#L136

As you wish, I've added examples for absolute path. 

Just to clarify, this PR only fixes absolute paths in the include { ... } directive (config merger). It doesn’t change the subscription file://... handling.

MarksonHon (approved):

PR #980, opened 2026-04-23
fix: carry ipv4 udp fallback, outbound, trace and conn-state follow-ups

olicesx:
## Summary

This PR started from the IPv4 UDP reply path fix for source-port conflict cases, and now also carries the related follow-up changes that landed on the same branch during review and validation.

The main functional changes in this branch are:

- IPv4 UDP reply fallback for wildcard-bind port-conflict cases
- host-side sysctl restoration for packets re-entering from dae netns via `dae0`
- outbound dependency and dialer / DNS pooled-connection follow-ups
- trace command usability restoration
- conn-state unification and connection-state map sizing follow-ups
- shutdown / drain handling fixes and related test updates

## What changed

- add raw IPv4 UDP fallback for reply packets when the normal anyfrom socket path cannot bind or send
- apply the fallback before returning `ErrAnyfromBindFailed`, so negative-cache hits do not cause later replies to be dropped
- restore host-side IPv4 sysctls required by the UDP port-conflict workaround:
  - `net.ipv4.conf.dae0.rp_filter=0`
  - `net.ipv4.conf.all.rp_filter=0`
  - `net.ipv4.conf.dae0.arp_filter=0`
  - `net.ipv4.conf.all.arp_filter=0`
  - `net.ipv4.conf.dae0.accept_local=1`
- ignore expired egress return handoff entries during redirect-track publish
- warn when the interface callback queue is full instead of dropping events silently

## Outbound updates included

- update the `outbound` replacement dependency to newer AnyTLS / sticky-ip related revisions
- fix dialer latency bookkeeping so failed or synthetic health checks do not append misleading latency samples
- improve `NotifyLatencyChange` cleanup behavior
- clear request deadlines before reusing DNS UDP pooled connections

## Trace updates included

- restore trace command usability after the earlier branch changes
- adjust trace-side generated artifacts / cleanup targets to match the current trace build output
- add trace-side counters / plumbing updates used by the restored tracing path

## Conn-state and control-plane follow-ups included

- unify TCP / UDP connection-state handling into a single shared `conn_state` structure across userspace, eBPF maps, and related tests
- expose and document `bpf_conn_state_map_size` in config / example config
- increase the BPF connection-state map sizing defaults for better concurrency headroom
- adjust related control-plane / janitor / validation tests to the shared conn-state layout
- ensure connections are aborted correctly on drain timeout and cancellation during shutdown

## Other updates in this branch

- sync the example config with the new conn-state sizing and THP-related options
- carry the merged Arch Linux official repository installation documentation update already present on the branch

## Why

When dae needs to send an IPv4 UDP reply with a specific source port, the normal anyfrom path can fail if that port is already occupied by a wildcard bind on the host.

This mainly affects reply-style traffic such as DNS responses. Once the bind failure is negative-cached, later replies can keep failing even though a narrower fallback path would still work.

While validating and following up on this fix, the branch also accumulated the outbound, trace, conn-state, and shutdown-related changes listed above. This description is updated to reflect the actual current branch contents instead of only the original UDP fallback scope.

## Manual verification

- [ ] verify IPv4 UDP replies still work when the source port conflicts with an existing wildcard bind
- [ ] verify replies still work after anyfrom bind-negative-cache hits
- [ ] verify no regression on the normal UDP reply path
- [ ] verify outbound health checks do not preserve synthetic failure latency history
- [ ] verify reused DNS UDP pooled connections do not retain stale deadlines
- [ ] verify trace command works again on the updated branch
- [ ] verify conn-state related config / defaults still load and behave correctly
- [ ] verify shutdown / drain still aborts connections correctly on timeout or cancellation

Tested locally.

Manual verification completed:
- verified raw IPv4 UDP fallback delivers reply packets with the expected source address and port
- verified fallback still works after anyfrom bind-negative-cache hits
- verified expired egress return handoff entries are ignored safely
- verified dialer latency history is not preserved on synthetic health check failures
- verified DNS UDP pooled connections clear request deadlines before reuse
- verified trace command usability was restored on the branch
- verified conn-state related tests and config updates were aligned with the shared map layout
- verified shutdown coverage for drain timeout / cancellation was added and passes locally


MaurUppi:
等你这个 fix 合并后，我再来搞 metrics & endpoint 的 feature

MaurUppi:
@olicesx merged?

forsaken628:
I tested a related case that may not be covered by this PR yet.

  Config shape:

 ```
  global {
      lan_interface: eth0
      wan_interface: auto
      dial_mode: domain
  }

  dns {
      bind: "tcp+udp://192.168.3.2:53"

      upstream {
          alidns: "udp://223.5.5.5:53"
          googledns: "udp://192.168.3.1:53"
      }

      routing {
          request {
              qname(geosite:cn) -> alidns
              fallback: googledns
          }
      }
  }
```

  Observed behavior:

  - On the dae host, dig @192.168.3.2 example.com A works.
  - On another LAN host, dig @192.168.3.2 example.com A +time=2 +tries=1 times out.
  - tcpdump -ni eth0 'host <lan-client-ip> and port 53' on the dae host sees the query packet arrive:
    192.168.3.50.xxxxx > 192.168.3.2.53: A? example.com.
  - But the DNS listener log does not show the request, and no response is sent.

  The suspicious path is the LAN ingress DNS fast path in control/control_plane.go.

  When a LAN packet is destined to dae's own DNS listener address:

```go
  if realDst.Port() == 53 {
      if c.dnsListener != nil {
          listenAddr := c.dnsListener.Addr()
          if listenAddr != "" && listenAddr == realDst.String() {
              return
          }
      }
      ...
  }
```
  The comment says this should let the DNS listener handle the packet, but at this point the packet has already entered dae's LAN ingress userspace/tproxy path. Returning here
  appears to drop/ignore the packet instead of delivering it to the DNS listener socket or handling it through DnsController.

  So this is slightly different from a DNS reply-send failure:

  - the request reaches the LAN interface,
  - but it does not reach the DNS listener / DNS controller,
  - therefore the reply fallback path in this PR may not be exercised.

  This seems to affect configs where dns.bind listens directly on the LAN address, e.g. tcp+udp://192.168.3.2:53, and LAN clients use dae itself as their DNS server.

MarksonHon (approved):

PR #995, opened 2026-05-06
fix(bpf): fallback to bpf_get_current_comm when bpf_get_current_task is unavailable

### LINKED ISSUE #993: [Bug Report] eBPF load failure: bpf_get_current_task helper rejected in tproxy_wan_cg_connect4
### Checks

### Current Behavior

skip

### Steps to Reproduce

skip

### Description
`dae` fails to start due to a fatal eBPF verifier error during the object loading phase. The kernel verifier rejects the `tproxy_wan_cg_connect4` program because it attempts to use the `bpf_get_current_task` helper, which is not permitted for this specific BPF program type (likely `BPF_PROG_TYPE_CGROUP_SOCK_ADDR`).

### Environment
- **OS**: OpenWrt
- **Architecture**: arm64
- **Kernel Version**: 6.18.26 (with Netkit support)
- **dae Version**: `unstable-20260504.r841.320a9e`
- **Go Runtime**: `go1.26.2-X:heapminimum512kib,newinliner,simd linux/arm64`
- **BTF Setup**: Using external `vmlinux-btf` package for CO-RE compatibility (base OpenWrt kernel strips BTF).

### Logs
```text
[root@OpenWrt dae]# ./dae-linux-arm64 run -c config.dae 
[May 05 13:01:17]  INFO Include config files: [config.dae]
[May 05 13:01:17]  WARN so_mark_from_dae is unset; using internal socket mark 0x100 to prevent dae UDP self-capture
[May 05 13:01:17]  INFO Waiting for network...
[May 05 13:01:17]  INFO Network online.
[May 05 13:01:17]  INFO Building control plane and routing rules...
[May 05 13:01:17]  INFO Kernel v6.18.26 supports Netkit, attempting to create Netkit device pair
[May 05 13:01:17]  INFO Successfully created Netkit device pair dae0 <-> dae0peer via netlink API (scrub=false)
[May 05 13:01:17]  INFO Successfully created Netkit device pair dae0 <-> dae0peer using netlink API
[May 05 13:01:17]  INFO Successfully created Netkit device pair (performance mode)
[May 05 13:01:17]  INFO Loading eBPF programs and maps into the kernel...
[May 05 13:01:17]  INFO The loading process takes about 120MB free memory, which will be released after loading. Insufficient memory will cause loading failure.
[May 05 13:01:17]  INFO Safely enabled bpf_redirect_peer() (kernel v6.18.26, netkit+scrub=NONE)
[May 05 13:01:18] FATAL load eBPF objects: field TproxyWanCgConnect4: program tproxy_wan_cg_connect4: load program: invalid argument: program of this type cannot use helper bpf_get_current_task#35 (27 line(s) omitted)

### Anything else?

_No response_

### PR DESCRIPTION
**Pull request by QiuSimons**
**Background**
Commit `320a9e52` introduced the usage of `bpf_get_current_task` combined with `BPF_CORE_READ(task, mm, arg_start)` to extract a more accurate process name (full command-line arguments parsing) instead of relying solely on `bpf_get_current_comm`. 

**The Problem**
While this works on kernels that support the helper in all required hook types, not all kernels allow the use of `bpf_get_current_task` (helper #35) in `CGroupSockAddr` (`BPF_PROG_TYPE_CGROUP_SOCK_ADDR`) or `CGroupSock` programs. On some environments (e.g., specific OpenWrt or older kernels), this causes a fatal verifier load error:
```text
program tproxy_wan_cg_connect4: load program: invalid argument: program of this type cannot use helper bpf_get_current_task#35
```

**The Solution**
Following maintainer feedback, this PR implements a robust load-time check and fallback mechanism without parsing the kernel BTF directly in user space:

1. **Probe**: During BPF initialization (`control/bpf_utils.go`), we use `features.HaveProgramHelper(ebpf.CGroupSockAddr, asm.FnGetCurrentTask)` to check if the current kernel supports the helper for the required program types.
2. **Inject**: The probe result is injected into the BPF `.rodata` section as a constant (`HAS_BPF_GET_CURRENT_TASK`).
3. **Branch & DCE**: In `get_pid_pname`, we branch based on this constant. If `HAS_BPF_GET_CURRENT_TASK` is `0`, the logic falls back to `bpf_get_current_comm`. The BPF verifier performs Dead Code Elimination (DCE) on the unreachable branch, safely discarding the unsupported `bpf_get_current_task` calls and its associated CO-RE relocations (which are correctly poisoned by the `cilium/ebpf` loader).
4. **Logging**: If the fallback is triggered, a `WARN` log is emitted in user space so that the user is aware that process names might be less accurate or truncated.

This ensures the proxy runs smoothly on older/restricted environments while preserving the precise process name capabilities on modern kernels.

**Testing**
- Evaluated against `make ebpf && go test ./...` successfully.
- Updated `bpf_stub.go` and verified successful compilation with `dae_stub_ebpf` build tags.
- Verified that the BPF verifier successfully prunes dead branches ensuring no `invalid argument` load errors on unsupported systems.

Close: #993

### DIFF
--- a/control/kern/tproxy.c
+++ b/control/kern/tproxy.c
@@ -176,7 +176,7 @@ struct dae_param {
 	__u8 dae0peer_mac[6];
 	__u8 padding_after_mac[2]; // pad to align use_redirect_peer
 	__u8 use_redirect_peer;
-	__u8 padding1;
+	__u8 has_bpf_get_current_task;
 	__u16 padding2;
 	// dae_socket_mark is set on dae's own sockets (Anyfrom pool) to identify them.
 	// When bpf_sk_lookup_* finds a socket, we check this mark to skip dae's own sockets.
@@ -3304,6 +3304,17 @@ static int __noinline get_real_comm_loop_cb(__u32 index, void *data)
 static __always_inline int get_pid_pname(struct pid_pname *pid_pname)
 {
 	int ret;
+
+	// Populate tgid and timestamp first
+	pid_pname->last_seen_ns = bpf_ktime_get_ns();
+	pid_pname->pid = bpf_get_current_pid_tgid() >> 32;
+
+	if (!PARAM.has_bpf_get_current_task) {
+		if (bpf_get_current_comm(&pid_pname->pname, sizeof(pid_pname->pname)))
+			pid_pname->pname[0] = '\0';
+		return 0;
+	}
+
 	// Get pointer to args string.
 	struct task_struct *task = (void *)bpf_get_current_task();
 	char *args = (void *)BPF_CORE_READ(task, mm, arg_start);
@@ -3337,9 +3348,6 @@ static __always_inline int get_pid_pname(struct pid_pname *pid_pname)
 		}
 	}
 
-	// Populate tgid
-	pid_pname->last_seen_ns = bpf_ktime_get_ns();
-	pid_pname->pid = bpf_get_current_pid_tgid() >> 32;
 	return 0;
 }
 
--- a/control/bpf_stub.go
+++ b/control/bpf_stub.go
@@ -26,17 +26,17 @@ var errBpfObjectsUnavailable = errors.New("eBPF objects are unavailable in this 
 // Only safe with: (1) netkit device + scrub=NONE, (2) kernel >= 6.8 (CVE-2025-37959 fix).
 // When enabled, provides ~50% throughput improvement by bypassing CPU backlog.
 type bpfDaeParam struct {
-	_               structs.HostLayout
-	TproxyPort      uint32
-	ControlPlanePid uint32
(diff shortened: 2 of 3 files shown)


control/kern/tproxy.c
```diff
@@ -188,6 +188,7 @@ struct dae_param {
  * This ensures the variable is placed in .rodata section and
  * can be rewritten from userspace via RewriteConstants. */
 const volatile struct dae_param PARAM = {};
+const volatile __u8 HAS_BPF_GET_CURRENT_TASK;
```
jschwinger233: 这个放到上面的 PARAM 里吧

jschwinger233 (changes requested):
注入的常量放到 struct dae_param PARAM 里，其他的都挺好

jschwinger233 (approved):

PR #997, opened 2026-05-10
fix(control): preserve DNS state for group override options

ayanamists:
## What this PR does

When a group overrides health-check options, dae rebuilds a `GlobalOption` for cloned dialers. That rebuilt option preserved the transport cache namespace at the call site, but it did not preserve runtime state injected into the base option, especially `DaeDNS`.

As a result, cloned dialers in groups with `tcp_check_url`, `udp_check_dns`, `check_interval`, or `check_tolerance` overrides could skip `DaeDNS.WrapNodeDialer(...)` and fall back to the default resolver for node hostnames.

This patch passes the base option into `ParseGroupOverrideOption` and preserves the runtime fields needed by cloned dialers.

## Tests

- `go test -tags dae_stub_ebpf ./control -run TestParseGroupOverrideOptionPreservesRuntimeDialerState -count=1`
- `go test -tags dae_stub_ebpf ./control -count=1`

### DIFF
--- a/control/control_plane.go
+++ b/control/control_plane.go
@@ -585,10 +585,9 @@ func newControlPlaneWithContextOptions(
 				log.Debugln("\t<Empty>")
 			}
 		}
-		groupOption, err := ParseGroupOverrideOption(group, *global, log)
+		groupOption, err := ParseGroupOverrideOption(group, *global, log, option)
 		finalOption := option
 		if err == nil && groupOption != nil {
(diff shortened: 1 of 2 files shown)


jschwinger233 (approved):

jschwinger233:
貌似没有大问题，不过我不是特别熟悉 dns 这一部分，其他人有空的话也可以帮忙看一眼。

PR #1008, opened 2026-06-13
feat: add fixed_fallback policy — fixed() with health-aware failover

### LINKED ISSUE #1005: fixed(n) policy documentation: behavior when dialer is dead is unclear
The comment in `example.dae` says:

```
# Select the first node from the group for every connection.
#policy: fixed(0)
```

This is misleading. By reading the source code in `component/outbound/dialer_group.go`, `fixed(n)` does NOT check alive state at all — it always returns `g.Dialers[n]` regardless of whether that dialer is dead or alive.

Also, `policyNeedsAliveState()` returns `false` for `fixed`, which means no `AliveDialerSet` is even built for a fixed-policy group.

**Suggested comment improvement for `example.dae`:**

```
# Always select the n-th dialer from the group, regardless of alive state.
# No alive check, no fallback. If the dialer is in backoff, connections will fail.
#policy: fixed(0)
```

This would save users from confusion like "why is my traffic still going to a dead node" or "why does not fixed(0) switch away when the node is down".

Related source: `component/outbound/dialer_group.go` `_select()` method, `case consts.DialerSelectionPolicy_Fixed`.

itoywh:
## What

Adds a new dialer selection policy `fixed_fallback(n)` that behaves like `fixed(n)` but with health-aware automatic failover.

## Motivation

Currently, `fixed(n)` ignores alive state entirely — it always returns the n-th dialer regardless of whether the node is dead. This is documented in #1005 where the misleading comment in `example.dae` was discussed.

For users who want to pin traffic to a preferred node (e.g., a low-latency local node) while still having a safety net when that node goes down, there is no built-in solution.

## How It Works

`fixed_fallback(n)` (config syntax: `policy: fixed_fallback(0)`):

1. **When the fixed dialer is alive** → always select it (same as `fixed(n)`)
2. **When the fixed dialer is dead/backoff** → fall back to `min_moving_avg` among all alive dialers in the group
3. **When the fixed dialer revives** (periodic health check restores it) → traffic automatically returns to it

## Implementation

- `common/consts/dialer.go` — new constant `DialerSelectionPolicy_FixedWithFallback`
- `component/outbound/dialer_selection_policy.go` — parse `fixed_fallback(n)` syntax
- `component/outbound/dialer_group.go`:
  - `_select()` — new case: check fixed dialer alive → use it, else fallback with `min_moving_avg`
  - `policyNeedsAliveState()` — returns `true` for `FixedWithFallback` (needs AliveDialerSet)
- `example.dae` — document the new policy

Unlike `fixed()` which creates no `AliveDialerSet`, `fixed_fallback()` participates fully in the health check lifecycle, including automatic `AliveDialerSet` creation/destruction on `SetSelectionPolicy()` transitions.

## Related

- Closes #1005 (or at least provides the feature users expected from `fixed()`)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -74,7 +74,8 @@ jobs:
       - name: Extract dae binary
         run: |
           unzip -o /tmp/dae_artifact/dae-linux-x86_64.zip -d /tmp/dae_artifact/
-          ls -lh /tmp/dae_artifact/dae-linux-x86_64
+          ls -lh /tmp/dae_artifact/build/dae-linux-x86_64
+          mv /tmp/dae_artifact/build/dae-linux-x86_64 /tmp/dae_artifact/dae-linux-x86_64
 
       - name: Update GitHub Release (delete old + upload new + update body)
         env:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -53,3 +53,32 @@ jobs:
       check-run-id: "dae-bot[bot]/main-build-passed"
       check-run-conclusion: ${{ needs.build.result }}
     secrets: inherit
+
+  update-release:
+    if: needs.build.result == 'success'
+    needs: [build]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout codebase
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Download x86_64 artifact
+        uses: actions/download-artifact@v4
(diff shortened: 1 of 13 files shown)


itoywh:
Superseded by #1009 which adds configurable fallback strategy, timeout/retry control, and supports all 4 official dialer policies (`random`, `min_moving_avg`, `min`, `min_avg10`). Closing in favor of the enhanced implementation.

Issue #1029, opened 2026-06-24
UDP data path: a down node can remain selectable until ~50 traffic failures, and the retry-limit warning lacks diagnostics (related to #1026)


MaurUppi:
This is split out from a comment I made on #1026. After a static code review of the
connectivity-check model, I now believe what I observed is a **distinct issue** that is
*related to* but **not the same root cause as** #1026. Filing it separately for clarity.

## Code baseline

I observed this on a fork branch whose tree is:

> upstream `dae/main` (`5a51cc7`, "#1010") **+ PR#1011** (dynamic CheckOpts, still open) **+ PR#1015** (metrics endpoint, still open).

All mechanisms cited below were **verified to exist in upstream `dae/main` @ `5a51cc7`**, so this
is not fork-specific. Line numbers are upstream `dae/main`. One caveat: PR#1011's
"skip unconfigured probes" can make `checkUnused()` exit more readily, which may *widen*
exposure to the lifecycle gap discussed in #1026 — but it is not the cause of the behavior below.

## Summary

A proxy node that is effectively down can keep being **selected for UDP data traffic**,
producing repeated `WriteTo()` failures that surface only as:

```text
level=warning msg="Touch max retry limit." dialer="[node]" network=udp4 retry=3 src="[redacted]"
level=warning msg="handlePkt: touch max retry limit"
```

Observed without any reload/restart, across multiple UDP flows over several seconds.

## Why this is separate from #1026

#1026 is about the health-check **goroutine lifecycle**: `RegisterAliveDialerSet()`
(`component/outbound/dialer/connectivity_check.go:849`) restores the reference count but does
not restart an exited `aliveBackground()` (`:467`), so a re-referenced dialer can keep stale
**DNS-UDP / TCP** health state.

But the symptom above is on the **UDP _data_ path** (`network=udp4`, `IsDns:false`;
`control/udp.go:888` sets `UdpHealthDomain: dialer.UdpHealthDomainData`), and that path is
**decoupled** from `aliveBackground()`:

- **UDP data selection reads the data-UDP collection (`IdxUdp4/6`), not DNS-UDP.**
  `StandardHealthKeys()` (`health_domain.go:135`) enumerates a `HealthDomainDataUDP` key →
  `CollectionIndex()` maps it to `IdxUdp4/6` (`health_domain.go:64,83`);
  `buildSelectionState()` creates a dedicated alive-set at that index
[...]

PR #1030, opened 2026-06-28
fix(udp): treat only n < len(b) as a short write (#1029)

MaurUppi:
## Problem

On the UDP data path, `UdpEndpoint.WriteTo` (`control/udp_endpoint_pool.go:839`) treated any `n != len(b)` as a short write — retiring the endpoint and returning `io.ErrShortWrite`. But a proxy `WriteTo` reports the **on-wire** byte count, which includes the outbound protocol's encapsulation overhead (e.g. shadowsocks AES-128-GCM + IPv4 = +39 bytes), so `n` can legitimately exceed `len(b)` on a fully successful write. UDP datagrams are atomic; only `n < len(b)` is a genuine short write.

The over-count was misclassified as a short write, retiring healthy endpoints and exhausting the retry limit — surfacing as the `Touch max retry limit.` warnings reported in #1029. This `n != len(b)` line is present in upstream `dae/main` (this change is a clean cherry-pick onto it).

## Evidence

This was found by instrumenting the (previously debug-level) `WriteTo` errors that #1029 noted were "not captured at the time." With diagnostics enabled, **10655/10655** retry-exhaustion drops were exactly `n = len(b)+39` over-writes — datagrams delivered on the wire — with **zero** genuine under-writes and no `err != nil` failures. An over-count, not an error return, is the signature of a falsely-retired healthy endpoint rather than a down node.

After deploying `n < len(b)`, a 22.3h production window logged **zero** `Touch max retry limit.` events.

## Scope

Correctness fix only, plus two regression tests:

- `TestUdpEndpointWriteTo_AcceptsProxyOverheadBytes` — a `+39` over-write stays alive.
- `TestUdpEndpointWriteTo_StillRetiresOnTruePartialWrite` — a genuine `n < len(b)` still retires the endpoint and returns `io.ErrShortWrite`.

This addresses the **retry-limit-warning** aspect of #1029. The issue's design question (data-UDP alive defaults to `true` + 50-failure traffic threshold; whether to actively probe the data-UDP path) concerns the genuinely-down-node path (`err != nil`, evaluated *before* this check) and remains open. The observability improvement (#1029 suggestion 1) is intentionally kept out of this PR to keep it minimal.

### DIFF
--- a/control/udp_endpoint_pool.go
+++ b/control/udp_endpoint_pool.go
@@ -839,7 +839,11 @@ func (ue *UdpEndpoint) WriteTo(b []byte, addr string) (int, error) {
 		return n, err
 	}
 	ue.hasSent.Store(true)
-	if n != len(b) {
+	// A proxy WriteTo may report the on-wire byte count, which includes the
+	// protocol's encapsulation overhead (e.g. shadowsocks AES-128-GCM + IPv4 adds
+	// 39 bytes), so n can legitimately exceed len(b). UDP datagrams are atomic, so
+	// only n < len(b) is a genuine short write worth retiring the endpoint over.
+	if n < len(b) {
 		ue.retire()
 		return n, fmt.Errorf("%w: udp endpoint wrote %d/%d bytes to %s", io.ErrShortWrite, n, len(b), addr)
 	}
--- a/control/udp_endpoint_pool_test.go
+++ b/control/udp_endpoint_pool_test.go
@@ -1978,3 +1978,56 @@ func TestUdpEndpointPoolGetOrCreate_NoReplyTimeoutReleasesPrewarmedAnyfrom(t *te
 		return countPooledAnyfromConns(DefaultAnyfromPool) == 0
 	})
 }
+
+// #1029: a proxy WriteTo may return the on-wire byte count (payload plus the
+// protocol's encapsulation overhead, e.g. shadowsocks AES-128-GCM + IPv4 = +39),
+// which is strictly greater than len(b). That is a SUCCESSFUL write, not a short
+// write, and must not retire the endpoint.
+func TestUdpEndpointWriteTo_AcceptsProxyOverheadBytes(t *testing.T) {
+	const shadowsocksIPv4Overhead = 39 // SaltLen 16 + ATYP/IPv4/port 7 + TagLen 16
+	payload := make([]byte, 1250)
+	conn := &scriptedPacketConn{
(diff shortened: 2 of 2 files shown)


itoywh:
Independently ran into this exact bug on a production side-router and traced it to the same line (`n != len(b)` in `control/udp_endpoint_pool.go`) before finding this PR. Confirming your fix direction and adding a few extra data points that may help review (opened #1055, now closed as a duplicate of #1029):

**1. Also reproduces with AES-256-GCM (+55 bytes), not just AES-128-GCM (+39).**
Real log line from a falsely-retired endpoint:
```
udp endpoint wrote 1305/1250 bytes
```
`1305 - 1250 = 55`, which is exactly the shadowsocks `aes-256-gcm` UDP encapsulation overhead:
- salt: 32
- IPv4 address metadata (ATYP 1 + addr 4 + port 2): 7
- GCM auth tag: 16
- **total: 55**

So the over-count magnitude scales with the cipher's salt/tag size; `+39` (aes-128-gcm) and `+55` (aes-256-gcm) are both just "on-wire length > payload length", exactly as this PR describes.

**2. VMess (packet-addr) is affected too, not only shadowsocks.**
Same root cause: the proxy `WriteTo` returns the on-wire byte count. Quick comparison across protocols in the outbound lib:
- `trojanc/packet.go` → `return len(p), nil` — payload length, **not** affected
- `shadowsocks/udp_conn.go` → `return c.PacketConn.WriteTo(toWrite, ...)` — on-wire length, affected
- `vmess/packet.go` → `return c.write(buf)` where `buf = packetAddr + payload` — on-wire length, **affected**

So any encapsulating outbound whose `WriteTo` reports on-wire length will trip the old `n != len(b)` check.

**3. Real-world trigger: large QUIC datagrams.**
On my setup the false retirements were driven exclusively by large QUIC UDP writes (e.g. `play.googleapis.com:443`, ~1250-byte payloads). Small UDP (DNS) and all TCP were never affected, because they rarely produce a payload large enough to make the over-count noticeable — but the over-count is present on every successful write regardless of size; large datagrams just made the misclassification frequent enough to exhaust the retry limit and mark the node DEAD on udp4, followed by periodic-health-check "resuscitation" ~5 min later. Classic flap loop.

Net: your `n < len(b)` fix (plus normalizing the success return to `len(b)`) resolves all of the above. Thanks for the clean minimal patch — hope this helps it land.

Issue #1037, opened 2026-07-09
Reload inherits stale health-check state and defers re-probe, so changes to tcp_check_url/udp_check_dns/check_interval are deferred by one full check cycle when all nodes are ALIVE


itoywh:
### Summary
When `tcp_check_url`, `udp_check_dns`, or `check_interval` are changed and applied via `reload` (SIGHUP / `dae reload`), the new health-check configuration takes effect only after a delay. The previous alive/dead verdict is inherited, and the first re-probe of the new target is deferred by one full `check_interval` cycle. As a result, the config change appears to be ignored unless a full `restart` is performed.

### Expected behavior
Changing the health-check target (e.g. pointing `udp_check_dns` at a different or unreachable resolver) and reloading should re-evaluate node health against the new target, or at least begin probing it promptly.

### Actual behavior
After `reload`:
- Nodes that were `alive` before the reload stay `alive` (the old snapshot is restored).
- If **all inherited collections are ALIVE**, `reloadInheritedHealth` is set and the first probe of the (possibly changed) target is **deferred by one full `check_interval` cycle** (e.g. 300s). During this window the config change appears to have no effect.
- If any collection was **DEAD** before the reload, the replacement dialer probes immediately (no deferral).

In our testing, changing `udp_check_dns` to an unreachable address and reloading kept the node `alive` with **0** probe attempts against the new target; a full `restart` immediately produced timeouts against the new target.

### Root cause
On reload, `ControlPlane.InheritDialerHealthFrom(previous)` copies the previous generation's health snapshot into the new dialers via `d.RestoreHealthSnapshot(oldDialer.ReloadHealthSnapshot())` (`control/control_plane.go`). It matches dialers purely by group+name and does **not** compare whether the health-check *configuration* (url/dns/interval) actually changed.

Additionally, in `component/outbound/dialer/connectivity_check.go`, when `d.reloadInheritedHealth.Swap(false)` is true and `cycle > 0`, the initial probe delay is increased by a full `cycle`:

```go
if d.reloadInheritedHealth.Swap(false) && cycle > 0 {
    initialDelay += cycle
}
```

So even after the inherited snapshot is restored, the first real probe of the (possibly changed) target is postponed by `check_interval`.

The reload path is designed to preserve connection liveness for *unchanged* dialers (which is desirable to avoid traffic flaps), but it has the unintended side effect that **health-check config changes are invisible until the next periodic cycle**, and the inherited "alive" verdict can mask a now-broken check target.

[...]

PR #1016, opened 2026-06-16
fix(reload): tiered reload to eliminate BPF map races during staged handoff

olicesx:
## Summary

Fixes silent traffic blocking (`Operation not permitted`) after staged same-port hot handoff reload, reported in #1013.

### Root Cause

When a user modifies DNS runtime parameters (e.g., `optimistic_cache`, `optimistic_cache_ttl`, `max_cache_size`) and triggers a reload, the staged handoff path shares the BPF object between old and new control planes. The DNS fingerprint included these runtime-tunable fields, causing `domain_routing_map` to be **cleared synchronously** and **replayed asynchronously** (via a capacity-1024 channel with silent drops) while the old control plane was still actively routing packets on the shared map. This created a race window where the old plane lost domain routing entries, causing transient routing failures.

### Two-Layer Fix

**Layer 0 — DNS fingerprint fix:**
- Exclude runtime-tunable parameters (`OptimisticCache`, `OptimisticCacheTtl`, `MaxCacheSize`) from `dnsConfigFingerprint`; they are atomic-tunable via `DnsController.UpdateRuntime` and do not affect BPF map state.
- Add `dnsRoutingUnchanged` flag to `ControlPlane` (passed via constructor); when true, `CommitPreparedDatapath` and the non-deferred commit path skip `clearReloadDomainRoutingMap` + `replayDnsReloadCache`.

**Layer 1 — Tiered reload via `bpfDatapathChanged()`:**
- New function detects all B-class map changes: routing rules, routing fallback, group definitions, DNS routing, interface bindings, conn-state map size, and `SoMarkFromDae`.
- When BPF datapath config changes, force a **full reload** (fresh BPF objects) instead of staged handoff, eliminating the shared-map race window entirely.
- Safe (pure userspace) fields — `log_level`, check URLs, `dial_mode`, `sniffing_timeout`, TLS settings, bandwidth limits, `optimistic_cache`, etc. — continue to use staged handoff for zero-downtime reload.

### Reload Decision Matrix

| Config change type | Reload method | BPF maps |
|---|---|---|
| Runtime params (`optimistic_cache`, etc.) | Staged handoff | Shared, untouched |
| Safe params (`log_level`, check URLs, `dial_mode`, etc.) | Staged handoff | Shared, untouched |
| Routing / group / DNS routing changes | Full reload (new) | Fresh objects, no race |
| Port change | Full reload | Fresh objects |

### Test Coverage

[...]

PR #1039, opened 2026-07-09
feat(reload): self-heal TC filter attachment after interface binding (repair missing filters)

itoywh:
## Summary

Post-commit datapath **self-healing**: after `commitInterfaceBindings`, verify TC filters on `dae0`(handle 0x2022) + LAN/WAN(handle 0x2023) are attached; if any are missing, **re-attach them automatically** instead of just failing.

### Problem

`bindLan`/`bindWan` use `_=c.addQdisc(link)` (best-effort). Missing clsact qdisc causes subsequent `FilterAdd` to fail **silently** — errors logged but do not propagate to abort reload. Result: **reload reports success, daemon running, nodes alive, but user traffic bypasses proxy entirely**. Exact pattern from #1013, reproduced in production on 2026-07-09 (reload at 15:27 → silent breakage → restart at 16:30).

### Fix — `repairDatapathBindings` (self-heal)

New functions in `controlPlaneCore`:

- **`checkBinding(bi)`**: Verifies a single bound interface via `netlink.FilterList()` (same API as `PurgeStaleTCFilters`), returns `(ok, reason)`.
- **`missingBindings() ([]boundIface, fatal)`**: Returns any interfaces whose TC filter is missing. Only `dae0` is treated as fatal.
- **`repairDatapathBindings() ([]string, fatal)`**: Detects missing filters → re-runs `_bindLan`/`_bindWan` for the missing interfaces → single re-check → if recovered, log at **Debug** (silent success); if still missing, returns the item for the caller to surface as a **Warn/error**.
- `recordBoundIface` is de-duplicated (name+label+major) so re-attach never produces duplicate records.

Called at **both** commit paths:
1. Non-deferred constructor → catches startup failures immediately
2. Deferred `CommitPreparedDatapath` → self-heals reload failures before cutover

Unlike a pure validator, this does **not** abort on a missing filter — it repairs first, and only reports if repair fails. Successful self-heal is logged at Debug level to stay quiet on every restart.

(Legacy `validateDatapathBindings` is kept as a thin formatting wrapper over `missingBindings` for backward compatibility.)

### Test Plan

| Step | Expected |
|------|----------|
| Normal start | All filters present, no re-attach, no noise |
| Remove clsact qdisc during reload | Self-heal re-attaches missing TC filters; success logged at Debug |
| Filter truly unattachable | Reported as still-missing (Warn/error) |
| Reload unchanged config | Passes (regression guard) |
| Reload health-check only | Passes (isolated from #1038) |

Relates: #1013 #1016 #1037 #1038

### DIFF
--- a/control/control_plane.go
+++ b/control/control_plane.go
@@ -818,8 +818,10 @@ func newControlPlaneWithContextOptions(
 		// Validate that TC filters are actually attached.  A silent bind
 		// failure (e.g. missing clsact qdisc, interface disappeared) would
 		// otherwise cause traffic to bypass the proxy with no error.
-		if missing := core.validateDatapathBindings(); len(missing) > 0 {
-			return nil, fmt.Errorf("datapath validation failed after interface binding: %v", missing)
+		if core != nil {
+			if missing := core.validateDatapathBindings(); len(missing) > 0 {
+				return nil, fmt.Errorf("datapath validation failed after interface binding: %v", missing)
+			}
 		}
 		if plane.sharedBpfReload {
 			if err = clearReloadDomainRoutingMap(core.bpf.Load()); err != nil {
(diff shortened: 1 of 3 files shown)


itoywh:
@MaurUppi @olicesx

itoywh:
Closing in favor of #1048.

#1048 is a stronger, self-contained replacement for this PR:
- matches the **full** TC handle (this PR only compared the major 16 bits, so it could not distinguish the ingress/egress filters LAN/WAN share under major 0x2023);
- additionally verifies the attached **BPF program id** (`netlink.BpfFilter.Id`), catching 'zombie' filters (handle present but pointing to a stale program) that a handle-only check silently passes;
- is wired into the live startup/reload path (self-heal: missing LAN/WAN auto re-attached, missing dae0 fatal);
- degrades gracefully to a handle-only check when the program id cannot be resolved, so no regression into false negatives.

Relationship to #1047: #1048 covers handle-missing + program-mismatch zombies; #1047 covers the stale-flipped-handle zombie via make-before-break. Complementary, both can merge independently.

PR #1042, opened 2026-07-10
fix(dialer): mark node unavailable when a network type has no applicable IP

### LINKED ISSUE #1040: IPv6 (and IPv4) connectivity check never marks node dead when no A/AAAA record resolves (false-alive)
## Summary
When `tcp_check_url` resolves only to an A record (no AAAA), the **IPv6 TCP** connectivity-check collection is never marked dead. It stays at its initial `Alive=true` state forever, so a node that is actually unreachable over IPv6 is still treated as alive and can be selected for IPv6 traffic. The same applies symmetrically to the IPv4 collection when a domain has no A record.

## Root cause
In `component/outbound/dialer/connectivity_check.go`:

1. The TCP IPv6 `CheckFunc` (lines ~570-584) returns `(false, nil)` when `!opt.Ip6.IsValid()` (no AAAA record):
   ```go
   if !opt.Ip6.IsValid() {
       d.Log.WithFields(...).Debugln("Skip check due to no DNS record.")
       return false, nil
   }
   ```
   The IPv4 check has the same `!opt.Ip4.IsValid()` -> `return false, nil` pattern (lines ~553-559).

2. `Dialer.check()` (line 1260) handles the `(ok=false, err=nil)` case by breaking out of the retry loop (lines 1280-1283) and **neither** calls `markAvailable` (line 1286, `ok && err == nil`) **nor** `markUnavailable` (line 1324, `err != nil`):
   ```go
   if err == nil {
       // No applicable IP; skip.
       break
   }
   ```

3. The trailing comment at line 1348 documents this behaviour: `// Skip update when (ok=false, err=nil): preserve existing alive state.`

4. `newCollection()` initializes `Alive=true` (line ~134), so the affected collection is frozen at `Alive=true` and is never corrected.

## Impact
- Triggered when `tcp_check_url` is a DNS name whose resolution lacks a record for one IP family (most commonly: a domain with **no AAAA** record, used over an **IPv6** route).
- In that scenario, IPv6-targeted traffic may be routed to a node that is effectively unreachable over IPv6, with no health-check feedback - a dead path is reported as alive.
- Edge case / low frequency, but a correctness bug.

## Reproduction (simplified)
1. Set `tcp_check_url` to a hostname that returns only an A record (no AAAA).
2. At Debug log level, the IPv6 TCP check logs `Skip check due to no DNS record.`.
3. The IPv6 collection remains `Alive=true` indefinitely; `markUnavailable` is never called for it.

## Suggested fix
When the check target has no applicable IP for a given network type (`!opt.Ip6.IsValid()` for IPv6, `!opt.Ip4.IsValid()` for IPv4), explicitly `markUnavailable` for that network type (or otherwise treat "no record" as dead for the corresponding collection), instead of silently preserving the initial alive state.

A PR will follow once we verify the fix on our side.

## Note
This "skip when no record" logic appears to have been introduced in #970 (commit 306faa7e, "Enhance control plane features and improve CI workflows").

itoywh:
Closes #1040

## Summary
`Dialer.check()` returned `(ok=false, err=nil)` — i.e. "no applicable IP for this network type" (e.g. an IPv6 probe against a node that resolves to no AAAA record) — without touching the alive state. Because a `collection` defaults to `Alive=true`, such a node stayed falsely "alive" for a network it can never actually serve, so selection could keep routing that traffic to a dead path.

Explicitly call `markUnavailable` on the `(ok=false, err=nil)` branch so an inapplicable network type is never gets treated as alive.

## Test plan
- `go build ./...` and `go test ./component/outbound/...` (Linux)
- Verified on a veth-mode deployment: a node with no AAAA record no longer stays falsely alive on the IPv6 domain.

### DIFF
--- a/component/outbound/dialer/connectivity_check.go
+++ b/component/outbound/dialer/connectivity_check.go
@@ -37,6 +37,13 @@ import (
 
 const Timeout = 10 * time.Second
 
+// ErrNoApplicableIP is a sentinel error returned by CheckFunc when the target
+// has no DNS record for the requested IP version (e.g. no AAAA for IPv6).
+// It is distinguished from a plain (false, nil) skip so that check() can
+// mark the node unavailable for that network type without affecting the
+// general (ok=false, err=nil) "preserve alive state" contract used by tests.
+var ErrNoApplicableIP = stderrors.New("no applicable IP for this network type")
+
 type UdpHealthDomain uint8
 
 const (
@@ -489,7 +496,7 @@ func (d *Dialer) aliveBackground() {
 					"dialer":  d.property.Name,
 					"network": typ.String(),
 				}).Debugln("Skip check due to no DNS record.")
-				return false, nil
+				return false, ErrNoApplicableIP
 			}
 			return d.HttpCheck(ctx, IdxTcp4, opt.Url, opt.Ip4, opt.Method, tcpSomark, mptcp)
 		},
@@ -511,7 +518,7 @@ func (d *Dialer) aliveBackground() {
 					"dialer":  d.property.Name,
 					"network": typ.String(),
 				}).Debugln("Skip check due to no DNS record.")
-				return false, nil
+				return false, ErrNoApplicableIP
 			}
 			return d.HttpCheck(ctx, IdxTcp6, opt.Url, opt.Ip6, opt.Method, tcpSomark, mptcp)
 		},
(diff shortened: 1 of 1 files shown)


itoywh:
根因就是一行遗漏：`ControlPlane` 结构体缺少 `dnsRoutingUnchanged bool` 字段（#1016 refactor 时改了引用处但漏了结构体定义）。加了这一行，全部通过。

olicesx:
Heads-up: daeuniverse/dae#1099 (merged) already covers this — ErrNoApplicableIP + markUnavailable (same-source implementation) is already in #1099.

PR #1043, opened 2026-07-10
ci(release): fix missing '=' in pkg_arch assignment for i486/pacman

### LINKED ISSUE #1041: release.yml: missing `=` in `pkg_arch'i486'` breaks 386/pacman packaging
## Bug
In `.github/workflows/release.yml` (and `prerelease.yml`) line 174, the pacman branch of the `pkg_arch` case statement reads:

```yaml
'pacman') pkg_arch'i486' ;;
```

This is a shell assignment with a **missing `=`** — it should be `pkg_arch='i486'`.

## Impact
- The `i486` / `386` Arch Linux (`pacman`) packaging variant is mis-assigned: `pkg_arch'i486'` is not a valid assignment, so `pkg_arch` keeps its prior value and the `i486` target is effectively lost / the step parses incorrectly.
- Common arches (x86_64 / arm64 / deb / rpm) are unaffected, so daily releases work — but the 386 pacman artifact is broken.
- Upstream-inherited, present since v2.0.0 final (`fee4c866`), also still present on `main`.

## Suggested fix
```yaml
'pacman') pkg_arch='i486' ;;
```

## Note
Found while reviewing the release workflow in fork `itoywh/dae`. We don't build 386 so it doesn't block us, but worth fixing upstream. Happy to open a PR.

itoywh:
Closes #1041

## Summary
In the `'386'` arch case of the package-build step, the `pacman` branch wrote `pkg_arch'i486'` **without the `=` sign**, so the shell parsed it as a command named `pkg_arch` with argument `i486` instead of an assignment. The pacman target architecture was therefore never set, and i486/386 (`pkg_arch='i486'`) packages would be built with a wrong/empty arch.

Restore the missing `=` so the assignment takes effect.

This is an upstream bug present in v2.0.0 final (`fee4c866`); both `release.yml` and `prerelease.yml` are fixed.

## Test plan
- `grep -n "pkg_arch'i486'" .github/workflows/release.yml .github/workflows/prerelease.yml` now returns nothing.
- Only affects 386 Arch/pacman packaging (we do not build 386 in daily CI), so no runtime impact on other arches.

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -171,7 +171,7 @@ jobs:
                 case $pkg_mgr in
                   'deb') pkg_arch='i386' ;;
                   'rpm') pkg_arch='x86' ;;
-                  'pacman') pkg_arch'i486' ;;
+                  'pacman') pkg_arch='i486' ;;
                 esac ;;
               'arm64')
                 if [ $pkg_mgr == 'deb' ]; then
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -171,7 +171,7 @@ jobs:
                 case $pkg_mgr in
                   'deb') pkg_arch='i386' ;;
                   'rpm') pkg_arch='x86' ;;
-                  'pacman') pkg_arch'i486' ;;
+                  'pacman') pkg_arch='i486' ;;
                 esac ;;
               'arm64')
                 if [ $pkg_mgr == 'deb' ]; then


olicesx:
Closing keyword  was included in the merged #1099 body (architectural note: superseded by the TCX/netns rework there). This PR is redundant now and can be closed — flagging for maintainers since the author cannot close others' PRs. cc @MarksonHon

olicesx:
Heads-up: daeuniverse/dae#1099 (merged) already covers this — release.yml / prerelease.yml `pkg_arch'i486'` missing `=` is fixed in #1099 (shellcheck-verified against the upstream baseline).

PR #1044, opened 2026-07-11
fix(control): guard core.Close() in closeTail with recover()

itoywh:
## Summary
`closeTail()` is invoked from an async goroutine — the `go func() { done <- c.closeTail() }()` caller in `shutdownAfterSignalWithHandoff`, which enforces a 5s `controlPlaneDeferredCleanupTimeout`.

`core.Close()` performs netlink/BPF teardown and can panic on rare kernel/state edge cases. Because `closeTail()` runs inside that goroutine, an **unrecovered panic** there crashes the goroutine silently and only surfaces as an opaque 5s timeout — the real root cause is lost, making shutdown/reload failures hard to diagnose.

## Change
Wrap `core.Close()` in a `recover()` so the cleanup path stays best-effort: on panic we log `[Close] core.Close panicked (recovered): <reason>` and continue, instead of crashing the goroutine.

## Scope / notes
- Minimal, self-contained change to `closeTail()` only.
- No behavior change on the happy path; only affects the (rare) panic path.
- Targets the synchronous `core.Close()` call already on `main`; does not depend on any staged-handoff refactor.

## Test plan
- [ ] `go build ./...` (existing CI)
- Manual: trigger a reload and confirm no regression in shutdown logs; the recovered-panic branch is exercised only on an actual `core.Close()` panic (hard to trigger intentionally, defensive only).

### DIFF
--- a/control/control_plane.go
+++ b/control/control_plane.go
@@ -3578,11 +3578,26 @@ func (c *ControlPlane) closeTail() error {
 
 	// Note: inConnections is cleared by AbortConnections() which should be called before Close()
 
-	// Combine defer errors with core.Close error
+	// Combine defer errors with core.Close error.
+	//
+	// Guard core.Close() with recover(): it performs netlink/BPF teardown and
+	// can panic on rare kernel/state edge cases. closeTail() is invoked from an
+	// async goroutine (see the `go func() { done <- c.closeTail() }()` caller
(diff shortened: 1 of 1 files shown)


olicesx:
Heads-up: daeuniverse/dae#1099 (merged) already covers this — the close path is synchronous in #1099 with a 5s backstop and retirement gate; the upstream recover() patch targets the old architecture.

PR #1048, opened 2026-07-13
control: program-aware datapath self-heal (supersedes #1039)

itoywh:
## Summary

This PR replaces the weak upstream #1039 with a stronger, self-contained datapath self-heal, and additionally layers in the **program-aware** check (the defense #1039 alone could not provide).

### What it does
- After `_bindLan` / `_bindWan` / `bindDaens` successfully attach a TC filter, the binding is recorded as `boundIface{ name, label, filters []boundFilter }`, where each `boundFilter` carries the **full** TC handle (`major<<16 | minor`) **and** the `*ebpf.Program` the filter is expected to point to. `dae0peer` (which lives inside the dae netns) is recorded too and validated there via `GetDaeNetns().WithRequired`.
- `commitInterfaceBindings()` resets the recorded set at the start of each run, then `NewControlPlane` / `CommitPreparedDatapath` call `repairDatapathBindings()` once after binding.
- `repairDatapathBindings()` auto re-attaches **missing LAN/WAN filters** (self-heal); a **missing `dae0` is fatal** and aborts startup/reload; `dae0peer` is surfaced as a non-fatal warning.

### Why stronger than #1039
- The weak #1039 only compared the **major** 16 bits of the handle. This matches the **full handle**, so it correctly distinguishes the two filters LAN/WAN carry under the same major (`0x2023…ingress` vs `0x2023…egress`).
- `hasDaeTcFilter` now also verifies the attached **BPF program id** (`netlink.BpfFilter.Id`, populated from `TCA_BPF_ID`) against the expected program. This catches **"zombie" filters** — a handle that is present but points to a stale / different / unloaded BPF program (e.g. after a BPF reload swapped the program under a persistently-attached handle) — which a handle-only check would silently pass. If the program id cannot be resolved (no `prog.Info()`) or the kernel does not report it, validation **degrades gracefully to a handle-only check**, so it never regresses into false negatives.

### Relationship to #1047
This PR and #1047 are complementary, not overlapping:
- This PR: mode ① (TC handle missing) + mode ③ (program-mismatch zombie).
- #1047: mode ② (stale flipped-handle zombie) via make-before-break deletion on reload.
Both can be merged independently.

### Tests
`control_plane_core_selfheal_test.go` adds table-driven coverage for `validateDatapathBindings`, `repairDatapathBindings`, and `hasDaeTcFilter` (handle-only, zombie rejection, healthy match, kernel-id-missing degradation, non-BPF-filter degradation). Program-aware cases use a real trivial eBPF program and skip gracefully where the kernel lacks support.

### Note
The self-heal path is wired into the live startup/reload flow (previously the helpers existed but were only exercised by unit tests). Please exercise on real hardware / in CI e2e before merge.

Supersedes #1039.

### DIFF
--- a/control/control_plane.go
+++ b/control/control_plane.go
@@ -815,6 +815,19 @@ func newControlPlaneWithContextOptions(
 		if err = plane.commitInterfaceBindings(); err != nil {
 			return nil, err
 		}
+		// Confirm TC filters are actually attached. A silent bind failure
+		// (e.g. missing clsact qdisc, interface disappeared) would otherwise
+		// cause traffic to bypass the proxy with no error. Missing LAN/WAN
+		// filters are auto re-attached (self-heal); a missing dae0 aborts.
+		if plane.core != nil {
+			if missing, fatal := plane.core.repairDatapathBindings(); len(missing) > 0 {
+				msg := fmt.Sprintf("datapath validation failed after interface binding (self-heal could not recover): %v", missing)
+				if fatal {
+					return nil, fmt.Errorf("%s", msg)
+				}
+				plane.log.Warnf("%s", msg)
+			}
+		}
 		if plane.sharedBpfReload {
 			if err = clearReloadDomainRoutingMap(core.bpf.Load()); err != nil {
 				return nil, fmt.Errorf("clearReloadDomainRoutingMap: %w", err)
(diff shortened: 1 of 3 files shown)


olicesx:
Heads-up: daeuniverse/dae#1099 (merged) already covers this — TCX links are attached with kernel anchoring in #1099; attach failure is a hard error, no silent clsact bypass to heal.

PR #1049, opened 2026-07-13
fix(component): re-bind interface when its operational state returns to Up

itoywh:
## Problem

The kernel removes the ingress qdisc and attached TC filters when a link goes down (admin down or carrier loss). On re-up they are not re-added automatically.

dae only (re)binds an interface's TC filters when the interface first appears (the `upLinks` gate in `InterfaceManager.monitor`). An admin-down -> up cycle does **not** emit a `RTM_DELLINK` — it only sends `RTM_NEWLINK` — so the binding callback was never re-fired. Result: after a flap, that interface silently loses all traffic steering until the next full daemon restart.

## Root cause

`component/interface_manager.go` `monitor()` skipped `RTM_NEWLINK` for any interface already recorded in `upLinks`, without considering the link's operational state. The bind callbacks (`_bindLan`/`_bindWan`, registered in `control_plane_core.go`) already know how to re-attach the filters; the gap was purely the event gate.

## Fix

Track each interface's last observed `OperState`. Re-fire `newCallback` when a known link transitions back to `OperUp` (in addition to the existing first-appearance path). The bind path is idempotent (`FilterDel` best-effort + `FilterAdd` tolerating `EEXIST`), so re-binding on a flap is safe and harmless.

## Test

`TestInterfaceManagerRebindsOnOperStateUpTransition` covers: first appearance binds, a repeated `OperUp` does not re-bind, `OperDown` does not bind, and the `OperDown -> OperUp` transition re-binds.

## Scope

Based directly on fee4c866 (v2.0.0). No unrelated changes.

### DIFF
--- a/component/interface_manager.go
+++ b/component/interface_manager.go
@@ -29,16 +29,23 @@ type InterfaceManager struct {
 	mu        sync.Mutex
 	callbacks []callbackSet
 	upLinks   map[string]bool
+	// operStates records the last observed operational state per interface so
+	// that we can detect a link coming back up after an admin/carrier flap and
+	// re-trigger the binding. The kernel removes the ingress qdisc and the
+	// attached TC filters when a link goes down, and they are not re-added
+	// automatically on re-up, which would silently break traffic steering.
+	operStates map[string]netlink.LinkOperState
 }
 
 func NewInterfaceManager(log *logrus.Logger) *InterfaceManager {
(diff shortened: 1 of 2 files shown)


olicesx:
Heads-up: daeuniverse/dae#1099 (merged) already covers this — link down→up rebinding is handled by the kernel-anchored TCX link model in #1099, no per-interface tc filter re-attach needed.

PR #1051, opened 2026-07-13
fix(cmd): re-queue reload request dropped during an in-flight reload

itoywh:
## Summary

`tryQueueReloadRequest` silently drops a reload signal whose `CompareAndSwap` on `reloadPending` loses because a reload is already queued or running. The dropped request is never retried, so the latest configuration change is lost until the next manual reload.

## Root cause

During a reload (staged handoff + retirement), `reloadPending` stays `true` until the old generation is fully retired. Any `dae reload` (SIGUSR1/SIGUSR2) arriving in that window hits the `CompareAndSwap(false, true)` failure branch and returns `false` with only a log warning — no compensation. After the in-flight reload finishes, the dropped request's intent (apply the newest config) is gone.

This is most visible when an automation/script issues `dae reload` faster than a reload completes, or when two quick manual reloads race.

## Fix

- Add a `reloadMissed` flag to `reloadManager`. The CAS-failure drop path now sets it.
- `finishReloadSuccess` / `finishReloadFailure` re-queue **exactly one** follow-up reload once `reloadPending` clears (post-retirement for the success path, synchronously for the failure path).

The re-queued request reads the current configuration, so a burst of dropped requests collapses into a single follow-up reload (further signals arriving during that follow-up are coalesced by the worker's `coalesceReloadRequest`). This keeps the system eventually consistent without an unbounded reload storm: once no more signals arrive, the last dropped request is honored and the chain stops.

## Testing

- `TestTryQueueReloadRequestRejectsConcurrentReload`: now asserts `reloadMissed` is set after a concurrent request is dropped.
- `TestReloadManagerRequeuesMissedRequest` (new): simulates an in-flight reload that drops a request, then verifies the `finishReload`-style `maybeRequeueMissedReload` re-queues exactly one normal reload and clears the flag.

Verified `gofmt` clean and `GOOS=linux go build -tags dae_stub_ebpf ./cmd/...` plus `go test -c` compile.

## Notes

This is independent of the datapath/self-heal reload work (#1047/#1048) and the zero-node rollback guard (#1050); it addresses a different hole (concurrent request loss at the signal layer).

### DIFF
--- a/cmd/reload_manager.go
+++ b/cmd/reload_manager.go
@@ -29,6 +29,8 @@ type reloadManager struct {
 	reloading                    atomic.Bool
 	reloadActive                 atomic.Bool
 	reloadPending                atomic.Bool
+	reloadMissed                 atomic.Bool
+	log                          *logrus.Logger
 	mu                           sync.Mutex
 	reloadingErr                 error
 	lastRetirementMu             sync.Mutex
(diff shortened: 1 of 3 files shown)


itoywh:
@olicesx

olicesx:
Heads-up: daeuniverse/dae#1099 (merged) already covers this — reload requests are never silently dropped in #1099 (reloadPending/reloadActive CAS + buffered request slot).

PR #1056, opened 2026-07-17
chore(ebpf): correct misleading comment regarding CO-RE relocation failure

KongQBin:
### Background

I am correcting a misjudgment I made in a previous commit. 
The previous comment incorrectly attributed the CO-RE bad relocation (stripping UAPI structs) to GCC 15 Dead Type Elimination (DTE). 
Recently, while debugging a separate LSM loading issue, I discovered that the root cause was actually related to the DWARF debug information level. 
This prompted me to go back and verify the original dae code, and it turned out to be the exact same issue. 
When the kernel is compiled with DWARF5, pahole(v1.31) may fail to parse it correctly, silently dropping UAPI structures during BTF generation. 
I am updating this comment to correct my previous assumption and to ensure future developers are not misled into investigating GCC compiler bugs. 
The actual workaround for the kernel is to compile with CONFIG_DEBUG_INFO_DWARF4=y.

### Checklist

### Full Changelogs

- Fix: Correct misleading code comment regarding CO-RE relocation failure and GCC 15 DTE.

### Issue Reference

N/A

### Test Result

Code compiles successfully. eBPF programs load without errors in the current environment.

### DIFF
--- a/control/kern/tproxy.c
+++ b/control/kern/tproxy.c
@@ -3,7 +3,13 @@
 
 // +build ignore
 
-// Disable implicit CO-RE from vmlinux.h to bypass bad relocation caused by GCC 15 DTE stripping UAPI structs.
+// Disable implicit CO-RE from vmlinux.h to bypass bad relocation.
+// Note: Previously misattributed to GCC 15 DTE. The actual root cause is that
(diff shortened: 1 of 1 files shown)


KongQBin:
@jschwinger233

jschwinger233 (approved):

PR #1070, opened 2026-08-08
fix(dns): honor must_rules in DNS fast path to restore bypass semantics

### LINKED ISSUE #1069: [Bug Report] must_rules bypass for DNS traffic broken since v2.0.0 (Must flag ignored in DNS fast path)
### Summary

`must_rules` no longer bypasses the DNS controller for DNS traffic (port 53) in v2.0.0. This is a regression introduced by PR #970, which added DNS fast-path handlers in the userspace UDP/TCP/ingress layers and removed the v1.x `Must` guard, but did not re-apply the `Must` check in the new fast paths.

### Environment

- **Dae version**: v2.0.0 (regression confirmed; v1.1.0 unaffected)
- **OS**: Linux
- **Kernel**: 7.x
- **Others**: WAN interface bound (localhost proxy mode)

### Current Behavior

In v1.x, the userspace UDP handler had this check (`control/udp.go:150`):

```go
if routingResult.Must > 0 {
    isDns = false // Regard as plain traffic.
}
```

This made `must_rules` work for DNS queries: when a process matched a `must_rules` rule (e.g. `pname(dig) -> must_rules`), its DNS queries (port 53) would bypass the DNS controller and be forwarded as plain traffic, respecting the `@server` argument of tools like `dig`.

PR #970 (`85a1fc3`, "Enhance control plane features and improve CI workflows") introduced "DNS fast path" handlers in `control/udp.go`, `control/tcp.go`, and `control/control_plane.go` for performance (avoid `UdpEndpoint` lookup per DNS query). These fast paths are gated only by `Port() == 53` and unconditionally call `DnsController.Handle_`, **without checking `routingResult.Must`**. The same PR removed the v1.x `if Must > 0 { isDns = false }` guard.

As a result, in v2.0.0, `must_rules` rules have no effect on DNS traffic — all port-53 traffic is hijacked into the DNS controller regardless of the `Must` flag set by eBPF routing.

### Steps to Reproduce

**1. Add a `must_rules` rule for a DNS tool in the routing section:**

```
routing {
  pname(mosdns) -> must_rules   # or pname(dig), pname(smartdns), etc.
  # ... other rules ...
  fallback: direct
}
```

**2. Query a CDN domain via two different DNS servers.** `download.microsoft.com` has geographic routing, so different DNS resolvers return IPs from entirely different network ranges:

```
dig @114.114.114.114 download.microsoft.com +short
dig @8.8.8.8 download.microsoft.com +short
```

**3. Compare the two answers.**

### Expected Behavior

| | v1.x (and patched v2.0.0) | v2.0.0 (broken) |
|---|---|---|
| `must_rules` for DNS | Works — `@server` is respected | **Broken** — `@server` is ignored |
| `dig @A` vs `dig @B` on a geo-routed CDN | Different IPs (each server's node) | **Both equal dae's own `dns.upstream` answer** |
| Root cause | `if Must > 0 { isDns = false }` skips DNS controller | Fast path `if Port()==53 { Handle_() }` ignores `Must` flag |

The bug signal: **two different `@server` values yield the identical answer** (which equals your dae `dns.upstream`'s answer), proving `@server` was silently ignored.

### Impact

Any `must_rules` configuration that should bypass the DNS controller for port-53 traffic is silently ineffective in v2.0.0. This affects:
- DNS loop prevention for local DNS proxies (`pname(mosdns, dnsmasq, smartdns, AdGuardHome) -> must_rules`)
- Diagnostic tools (`dig`, `nslookup`, `host`) with `pname(...) -> must_rules`

### Root Cause Analysis

The `Must` flag is set by the eBPF layer (`tproxy.c` → `ROUTE_STATE_MUST`) when a `must_rules` rule matches, and is correctly propagated to userspace via `routingResult.Must`. However, the three DNS fast-path entries introduced by #970 never read this flag:

- `control/udp.go` — `if realDst.Port() == 53 { ... Handle_() ... }`
- `control/tcp.go` — `if dst.Port() == 53 { handleTCPDnsFastPath(...) }`
- `control/control_plane.go` — DNS ingress fast path → `Handle_()`

The fix is to guard these entries with a `Must` check, restoring the v1.x semantics. A PR with the fix will follow.

luochen1990:
### Background

PR #970 introduced DNS fast paths in `control/udp.go`, `control/tcp.go`, and `control/control_plane.go` for performance, but they unconditionally route port-53 traffic into `DnsController.Handle_` without checking `routingResult.Must`. This breaks the `must_rules` bypass for DNS — see #1069 for full root cause analysis and reproduction steps.

### Fix

Guard all three fast-path entries with a `Must` check: when `routingResult.Must > 0`, skip the DNS controller and fall through to normal UDP/TCP forwarding.

- `control/udp.go`: wrap the fast-path body in `if routingResult.Must == 0 { ... }`
- `control/tcp.go`: same pattern
- `control/control_plane.go`: `if dnsRoutingResult.Must > 0 { goto skipDnsFastPath }` (goto avoids deep re-indentation of the heavily nested fast path)

This restores the v1.x semantics where `Must > 0` forces DNS traffic to be treated as plain traffic.

### Checklist

### Full Changelogs

- fix(control): DNS fast path now honors `routingResult.Must`, restoring `must_rules` bypass for DNS (regression from #970)

### Issue Reference

Closes #1069

### Test Result

**Reproduce config (minimal):**
```
routing {
  pname(mosdns) -> must_rules
  # ... other rules ...
  fallback: direct
}
```

**Verification via a geo-routed CDN domain** (`download.microsoft.com` returns IPs from entirely different network ranges depending on the resolver's location):

```
# Before fix: @server is ignored; both queries return the IDENTICAL IP
# from dae's own dns.upstream.
$ dig @114.114.114.114 download.microsoft.com +short
$ dig @8.8.8.8 download.microsoft.com +short   # ← identical to above

# After fix: @server is respected; different resolvers return their
# respective geographic nodes.
$ dig @114.114.114.114 download.microsoft.com +short   # → China-region IP
$ dig @8.8.8.8 download.microsoft.com +short           # → Akamai global IP
```

Bug signal: two different `@server` values yield the identical answer (dae upstream's), proving `@server` was ignored. Exact IPs vary by region/upstream; only the *equality* matters.

**Unit tests pass:**
```
$ go test -tags dae_stub_ebpf ./...
ok  	github.com/daeuniverse/dae/component/routing/domain_matcher	31.628s
ok  	github.com/daeuniverse/dae/control	(cached)
ok  	github.com/daeuniverse/dae/component/outbound/dialer	0.212s
... (all pass)
```

**eBPF tests:** no eBPF changes; `Must` is correctly populated via `RetrieveRoutingResult` — the break is only in userspace fast paths not reading it.

**gofmt clean:**
```
$ gofmt -l control/udp.go control/control_plane.go control/tcp.go
# (no output)
```

### DIFF
--- a/control/control_plane.go
+++ b/control/control_plane.go
@@ -3104,6 +3104,13 @@ func (c *ControlPlane) Serve(readyChan chan<- bool, listener *Listener) (err err
 							routingResult: dnsRoutingResult,
 						}
 
+						// Honor must_rules: if routing marked this flow as must, bypass the
+						// DNS controller and fall through to normal UDP forwarding.
+						// This restores the v1.x semantics where Must>0 forces isDns=false.
+						if dnsRoutingResult.Must > 0 {
+							goto skipDnsFastPath
+						}
+
 						dnsController := c.ActiveDnsController()
 						if dnsController == nil {
 							return
@@ -3165,6 +3172,7 @@ func (c *ControlPlane) Serve(readyChan chan<- bool, listener *Listener) (err err
 					}
 				}
 
+			skipDnsFastPath:
 				if !c.udpRouteScopeSensitive {
 					if ue, ok := DefaultUdpEndpointPool.Get(flowDecision.CachedRoutingEndpointKey()); ok {
 						if cached, cacheHit := ue.GetCachedRoutingResult(realDst, unix.IPPROTO_UDP); cacheHit {
--- a/control/tcp.go
+++ b/control/tcp.go
@@ -142,15 +142,20 @@ func (c *ControlPlane) handleConn(ctx context.Context, lConn net.Conn) (err erro
 	// Uses bufio.Reader to peek at data without consuming it,
 	// allowing proper fallback if this isn't DNS traffic.
 	if dst.Port() == 53 {
-		bufReader := bufio.NewReader(lConn)
-		handled, dnsErr := c.handleTCPDnsFastPath(ctx, lConn, bufReader, src, dst, routingResult)
-		if handled {
-			// Connection was handled as DNS - any errors are already logged
-			return dnsErr
+		// Honor must_rules: if routing marked this flow as must, bypass the
+		// DNS controller and fall through to normal TCP forwarding.
+		// This restores the v1.x semantics where Must>0 forces isDns=false.
+		if routingResult.Must == 0 {
+			bufReader := bufio.NewReader(lConn)
+			handled, dnsErr := c.handleTCPDnsFastPath(ctx, lConn, bufReader, src, dst, routingResult)
+			if handled {
+				// Connection was handled as DNS - any errors are already logged
+				return dnsErr
+			}
+			// Not DNS traffic (or failed to read as DNS) - fall through to normal TCP handling
+			// Wrap the connection to include buffered data that was peeked but not consumed
+			lConn = &bufioConn{Conn: lConn, reader: bufReader}
 		}
-		// Not DNS traffic (or failed to read as DNS) - fall through to normal TCP handling
-		// Wrap the connection to include buffered data that was peeked but not consumed
-		lConn = &bufioConn{Conn: lConn, reader: bufReader}
 	}
 
 	var (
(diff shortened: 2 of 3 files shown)


olicesx:
Heads-up: daeuniverse/dae#1099 (merged) already covers this — must-rules are honored inside the DNS fast path (TcpFlowBinding.Must) in #1099 — different mechanism, same guarantee.

PR #1081, opened 2026-08-16
fix(control): use DNS handoff when choosing dial target

### LINKED ISSUE #1054: [Bug Report] Panic in ChooseDialTarget during staged DNS controller handoff
## Checks

## Current Behavior

During a staged same-port reload, dae can panic while an old control-plane generation is still handling a TCP connection:

```text
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1026870]

goroutine 886759 [running]:
github.com/daeuniverse/dae/control.(*DnsController).HasDnsKnowledge(0x0, ...)
        github.com/daeuniverse/dae/control/dns_control.go:935 +0x50
github.com/daeuniverse/dae/control.(*ControlPlane).ChooseDialTarget(...)
        github.com/daeuniverse/dae/control/control_plane.go:1720 +0x16a
github.com/daeuniverse/dae/control.(*ControlPlane).chooseProxyDialer(...)
        github.com/daeuniverse/dae/control/dial.go:110 +0xbb
github.com/daeuniverse/dae/control.(*ControlPlane).routeDial(...)
        github.com/daeuniverse/dae/control/dial.go:234 +0xb4
github.com/daeuniverse/dae/control.(*ControlPlane).handleConn(...)
        github.com/daeuniverse/dae/control/tcp.go:233 +0xeb4
github.com/daeuniverse/dae/control.(*ControlPlane).Serve.func2.1(...)
        github.com/daeuniverse/dae/control/control_plane.go:3132 +0xb3
```

The panic happened about 55 ms after the staged generation entered `Serve` and about 5 ms after restoring the DNS reload cache:

```text
WARN Reload: Received reload signal; prepare to reload
WARN Reload: Prepare staged same-port handoff
WARN Reload: Serve
INFO Restored 80 DNS cache entries from previous control plane
panic: runtime error: invalid memory address or nil pointer dereference
```

This appears to be a DNS-controller ownership handoff race:

1. `reuseDNSControllerFrom` publishes the replacement controller through the old plane's handoff pointer.
2. It then sets `previous.dnsController = nil` and assigns the replacement to the new plane.
3. Old accepted/in-flight TCP work can still call `ChooseDialTarget` while the old generation is draining.
4. `ChooseDialTarget` reads `c.dnsController` directly instead of resolving the handoff with `c.ActiveDnsController()`.

The old generation therefore has a valid `dnsHandoffController`, but `c.dnsController` is nil. This direct access panics when `dial_mode: domain` reaches `HasDnsKnowledge`.

The same direct access and ownership-transfer sequence are present in the `v2.0.0` tag and current `main`:

```go
// control/control_plane.go
if c.dnsController.HasDnsKnowledge(c.dnsController.cacheKey(domain, ...)) {
```

```go
// control/dns_runtime.go
if publishHandoff != nil {
    publishHandoff(reusedController)
}
previous.dnsController = nil
r.dnsController = reusedController
```

## Expected Behavior

Connections still running on the old control-plane generation should resolve the DNS handoff controller and complete or drain without crashing the process.

`ChooseDialTarget` should take one stable `ActiveDnsController()` snapshot and handle nil safely, rather than directly reading the ownership field during reload.

## Steps to Reproduce

The race is timing-sensitive; a later reload with the same configuration and active traffic succeeded.

1. Configure dae with `dial_mode: domain` and a same-port reload configuration.
2. Keep TCP traffic active so the old control plane is accepting or handling connections during handoff.
3. Repeatedly run `dae reload` without changing the tproxy port.
4. Occasionally an old-generation connection reaches `ChooseDialTarget` after DNS ownership has transferred and dae panics as above.

A deterministic regression test could:

1. Create old and replacement control planes in domain mode.
2. Call `newCP.ReuseDNSControllerFrom(oldCP)` and verify `oldCP` resolves the published handoff while its owned `dnsController` is detached.
3. Invoke `oldCP.ChooseDialTarget(...)` for a non-reserved outbound and a non-IP domain.
4. Assert that it does not panic and uses the handoff controller.

## Environment

- **Dae version**: `unstable-20260710.r873.f920fa4` (downstream build based on upstream `v2.0.0`; the relevant code is unchanged from `v2.0.0`/current `main`)
- **Go runtime**: `go1.26.3-X:heapminimum512kib linux/amd64`
- **OS**: Debian GNU/Linux 13 (trixie)
- **Kernel**: `6.12.95+deb13-cloud-amd64`
- **Architecture**: x86_64
- **Relevant config**: `dial_mode: domain`; unchanged tproxy port, so reload used staged same-port handoff

## Anything else?

This is related to the staged handoff subsystem discussed in #1013/#1016, but appears to be a distinct userspace lifecycle race. #1016 addresses shared BPF/domain-routing-map mutation and does not change `ChooseDialTarget`, `ActiveDnsController`, or the ownership detachment shown above.

The crash is also operationally severe with the packaged systemd unit: a Go panic exits with status 2, while `Restart=on-abnormal` does not restart a process that exited normally with a non-zero status. That is secondary to the controller race, but it can turn this panic into a prolonged outage.

Zhanfg:
### Background

During DNS controller handoff, `ChooseDialTarget()` could still consult the old `c.dnsController` directly instead of the controller currently active through the handoff mechanism.

If DNS knowledge exists only in the handoff controller, the old lookup misses it and the dial target can incorrectly remain an IP address instead of switching back to the original domain.

This change uses `ActiveDnsController()` and handles the nil case before consulting DNS knowledge.

A regression test reproduces the handoff state and verifies that the domain is selected and rerouting is requested.

### Checklist

### Full Changelogs

- Use the active DNS handoff controller when deciding whether a dial target should use its original domain.
- Add a regression test covering DNS knowledge stored in the handoff controller.

### Issue Reference

Closes #1054.

### Test Result

- Added `TestChooseDialTargetUsesDNSHandoff`.
- Fork CI for the corrected regression fixture passed.
- Upstream CI is awaiting maintainer approval for fork-originated workflows.

### DIFF
--- a/control/control_plane.go
+++ b/control/control_plane.go
@@ -1617,7 +1617,8 @@ func (c *ControlPlane) ChooseDialTarget(outbound consts.OutboundIndex, dst netip
 			if isIPLikeDomain(domain) {
 				break
 			}
-			if c.dnsController.HasDnsKnowledge(c.dnsController.cacheKey(domain, common.AddrToDnsType(dst.Addr()))) {
+			dnsController := c.ActiveDnsController()
+			if dnsController != nil && dnsController.HasDnsKnowledge(dnsController.cacheKey(domain, common.AddrToDnsType(dst.Addr()))) {
 				// Has A/AAAA records. It is a real domain.
 				dialMode = consts.DialMode_Domain
 				shouldReroute = true
--- a/control/control_plane_dns_handoff_test.go
+++ b/control/control_plane_dns_handoff_test.go
@@ -1,0 +1,54 @@
+package control
+
+import (
+	"io"
+	"net/netip"
+	"testing"
+	"time"
+
+	"github.com/daeuniverse/dae/common/consts"
+	dnsmessage "github.com/miekg/dns"
+	"github.com/sirupsen/logrus"
+)
+
+func TestChooseDialTargetUsesDNSHandoff(t *testing.T) {
+	log := logrus.New()
+	log.SetOutput(io.Discard)
+
+	controller, err := NewDnsController(nil, &DnsControllerOption{Log: log})
+	if err != nil {
+		t.Fatalf("create DNS controller: %v", err)
+	}
+	defer func() { _ = controller.Close() }()
+
+	const domain = "example.com"
+	dst := netip.MustParseAddrPort("203.0.113.1:443")
+	controller.rememberDnsKnowledge(
+		controller.cacheKey(domain, dnsmessage.TypeA),
+		time.Now().Add(time.Minute),
+	)
+
+	controlPlane := &ControlPlane{
+		log: log,
+		controlPlaneGenerationState: controlPlaneGenerationState{
+			dialMode: consts.DialMode_Domain,
+		},
+	}
+	controlPlane.dnsHandoffController.Store(controller)
+
+	dialTarget, shouldReroute, dialIP := controlPlane.ChooseDialTarget(
+		consts.OutboundIndex(100),
+		dst,
+		domain,
+	)
+
+	if dialTarget != "example.com:443" {
+		t.Fatalf("dial target = %q, want %q", dialTarget, "example.com:443")
+	}
+	if !shouldReroute {
+		t.Fatal("expected handoff DNS knowledge to enable rerouting")
+	}
+	if dialIP {
+		t.Fatal("expected domain dial target, got IP dial mode")
+	}
+}


olicesx:
Heads-up: daeuniverse/dae#1099 (merged) already covers this — the DNS-handoff-aware dial target selection is already in #1099 (control_plane_dialtarget.go, HasDnsKnowledge).

PR #1100, opened 2026-09-16
docs: repair the examples and rewrite the manual in English and Chinese

Zakkaus:
### Background

Nine commands in the manual fail as written: two `tee` calls without `-a` keep
only the IPv6 forwarding line, `tee -a /etc/apt/sources.list/` names a
directory, `docker compose up` runs outside the clone, a backslash lands in the
plist DOCTYPE, `networksetup -setmanual` gets `/32` instead of a dotted mask,
`make CC=clang` sets a variable the Makefile never reads, the fish kernel check
calls `bat`, the CentOS 7 page installs from a repository section that
`mailbox.repo` ships disabled, and the macOS page pins Debian daily images that
have been removed. The routing page
does not mention two behaviour changes from the Unreleased changelog (named
parameters rejected, negated rules no longer merged). The quick start omits the
Dae Universe APT/RPM repository and the Nix flake. `docs/zh` has 8 of 21 pages.

The manual also described behaviour the code at `1ec85fe` does not have. The
rewrite corrects those passages against the code: sniffing is TCP-only under
`sniffing_timeout` and `dial_mode: ip` while QUIC on 443/8443 is always
sniffed; the 3-failure/10-minute negative cache is TCP-only; `dial_mode:
domain` re-matches routing once the domain is confirmed; kernel `domain()`
matching is keyed by destination IP, not by device; `asis` queries over UDP
and answers `TC=1` with an empty reply; `fixed_domain_ttl: 0` expires rather
than disables the cache; `ipversion_prefer` only filters when the preferred
family is known; `bootstrap_resolver` serves three lookup kinds, not only
startup; `disable_thp` always calls `prctl`; the tproxy listener lives in the
`daens` namespace, so the old firewall and `__UNROLL_ROUTE_LOOP` advice is
gone; the OPNsense 2000-byte DNS buffer limit no longer exists; and the
Vision and `cc_override` notes now match the pinned outbound library.

### Checklist

- [ ] There's an entry in the CHANGELOGS (documentation only)
- [x] There is a user-facing docs PR against https://github.com/daeuniverse/dae (this is it)

### Full Changelogs

- docs: fix examples that fail when run as written
- docs: record the two routing changes a config can trip over
- docs: split the routing reference into sections
- docs: rewrite the English manual and correct it against the code
- docs(readme): add the repository, Nix and an installation overview
- docs(zh): rewrite the Chinese manual and complete it to every English page

The English rewrite keeps every fenced block byte-identical; `contribute.md`
and `refactor-validation-plan.md` are untouched, `commit-msg-guide.md` only
gains a table. The negation note follows `component/routing/optimizer.go`, not
the changelog, which states the direction backwards.

### Issue Reference

CC: @olicesx @MarksonHon
Sync: https://github.com/daeuniverse/repo-for-linux/pull/17

### Test Result

- `markdownlint-cli2 'docs/**/*.md'`: 0 errors on each of the 6 commits.
- Repaired examples were executed: `printf | tee` writes both sysctl lines;
  the corrected plist passes `xmllint --noout`.
- Claims added to the routing and daemon pages were checked against
  `config/config.go`, `config/bootstrap_resolver.go`, `common/utils.go`,
  `control/tcp_sniff_policy.go`, `Makefile` and `install/dae.service` at `1ec85fe`.
- Each of the 28 corrected passages was checked by three independent code
  readings before the text was changed; two candidate corrections that failed
  that check were dropped.
- `mailbox.repo` lists `[mailbox-kernel]` with `enabled=0`;
  `kernel/mount-cgroup2.service` returns 404; the `bookworm/latest` images return 200.
- Every `docs/zh` page has the same heading sequence and fenced blocks as its
  `docs/en` counterpart (20 of 20).
- Not verified: the APT/RPM repository and Nix commands were not run on a
  fresh machine; they are what the manual site publishes and tests.


MarksonHon (approved):

Glassyiris (approved):
LGTM

PR #1114, opened 2026-09-19
fix(control): deliver question-less terminal DNS upstream replies

xunkutech:
### Background

A legitimate **terminal error** reply may omit the question section. DNSPod (119.29.29.29, the default direct/bootstrap resolver) answers `NXDOMAIN` with **QDCOUNT=0** for Bonjour/DNS-SD reverse lookups, e.g.

```
db._dns-sd._udp.0.5.168.192.in-addr.arpa. PTR
```

Apple devices on a LAN send that family every ~45 s (AirDrop/AirPlay, printers, Time Machine browsing). Two consequences:

1. In `resolveDNSUpstream` the RFC 5452 echo check (`questionEchoMatches`) sees no question and rejects the reply, so nothing is routed and the client cannot get an answer.
2. A question-less reply also cannot be matched back to the outstanding query, so a second dnsmasq upstream cannot rescue it either. Client-side the symptom is a plain **5 s timeout** — Bonjour service discovery is genuinely broken, not just noisier logs.

Reproduction (PTR + `in-addr.arpa` is required; an `A` query for a non-existent name gets a normal echo-bearing `NXDOMAIN` from the same server, which hides the bug):

```sh
dig @119.29.29.29 db._dns-sd._udp.0.5.168.192.in-addr.arpa PTR   # status: NXDOMAIN, QUERY: 0
dig @<router>      db._dns-sd._udp.0.5.168.192.in-addr.arpa PTR   # before: 5 s timeout
```

### Checklist

- [x] The Pull Request has been fully tested (unit test below + on-router A/B before/after)
- [x] There's an entry in the CHANGELOGS — release-time automation (`.github/workflows/generate-changelogs.yml`), nothing manual for this PR
- [x] There is a user-facing docs PR — not needed, this changes no config, API or user-facing behaviour other than answering correctly

### Full Changelogs

- [Implement] `control/dns_controller_handle.go`: restore the request question *before* the echo check, but only for the shape that cannot be an answer to some other query — **no question section, no records, and a non-success rcode**. Positive answers (rcode 0) and any reply that carries records must still echo the question, so the anti-spoofing guarantee is not weakened for anything that carries data.
- [Change] `component/dns/dns.go`: the response-routing failure path reports what actually arrived (`DNS response without question section (rcode=…, answers=…)`) instead of `qName cannot be empty`, which named an internal symptom rather than the upstream reply.
- [Implement] `control/dns_questionless_terminal_reply_test.go`: regression test pinning both halves — the three terminal rcodes (NXDOMAIN/REFUSED/SERVFAIL) must reach the client with the question restored, while a positive answer, or NOERROR/NODATA, without a question must still be dropped.

### Issue Reference

No issue was filed for this; the root-cause analysis is the Background section above.

### Test Result

```
$ go test -tags dae_stub_ebpf ./control/ -run TestQuestionlessTerminalUpstreamReplyIsDelivered -count=1 -v
--- PASS: TestQuestionlessTerminalUpstreamReplyIsDelivered (0.03s)
    --- PASS: .../nxdomain_without_question_is_delivered (0.01s)
    --- PASS: .../refused_without_question_is_delivered (0.00s)
    --- PASS: .../servfail_without_question_is_delivered (0.00s)
    --- PASS: .../positive_answer_without_question_is_still_dropped (0.00s)
    --- PASS: .../nodata_without_question_is_still_dropped (0.01s)
PASS
ok  github.com/daeuniverse/dae/control
```

The test is not vacuous: with the new guard disabled, the three "delivered" subtests fail, each with `reply does not echo the request question (possible spoofing or upstream cross-talk); dropped` — exactly the regression being prevented.

End-to-end on an affected x86_64 OpenWrt gateway, same deterministic trigger before/after:

| check | before | after |
|---|---|---|
| client `dig … in-addr.arpa PTR` | 5 s timeout | 15–17 ms `NXDOMAIN` |
| routing errors for that query family | +122 in 14.5 min | 0 in 60 s and in 3 min of natural traffic |
| A / TXT / AAAA / NXDOMAIN regression | 15/15 | 15/15 |

Honest scope note: in the Docker/OrbStack environment I used, the `control` package as a whole still reports `TestTCPOffloadSentAccountE2E` as failing (`no vdso address found in auxv` / `map create: operation not permitted`), which is an eBPF-in-container limitation — it fails identically on unpatched `main` there, and does not touch DNS.


xunkutech:
## Test evidence for this PR (revision 1c8803ba)

The CI workflows on this PR are still in `action_required` (first-time contributor; they need a maintainer to approve the run), so here are local results instead. Environment: `golang:1.26-bookworm` + clang-15/llvm-15 container on macOS, invoked with the documented command `go test -tags dae_stub_ebpf`.

### 1. Unit tests — whole module

```
$ go test -tags dae_stub_ebpf -count=1 ./...
```

21 packages pass. `control` is the only failing package, and inside it 433 test cases pass with exactly one failing test:

```
--- FAIL: TestTCPOffloadSentAccountE2E
    --- FAIL: TestTCPOffloadSentAccountE2E/IPv4
    --- FAIL: TestTCPOffloadSentAccountE2E/IPv6
    --- FAIL: TestTCPOffloadSentAccountE2E/IPv6DestinationOptions
    tcp_offload_e2e_verify_test.go:160: NewCollection (pruned): map .rodata: map create: operation not permitted (MEMLOCK may be too low, consider rlimit.RemoveMemlock)
```

That test loads the eBPF object, which cannot succeed in this container (with `--privileged` it fails earlier, at `no vdso address found in auxv`). It fails **identically on unpatched `main`** in the same environment (verified by stashing this patch and re-running), so it is environmental and unrelated to this change.

### 2. This change's own test

```
$ go test -tags dae_stub_ebpf ./control/ -run TestQuestionlessTerminalUpstreamReplyIsDelivered -count=1 -v
--- PASS: TestQuestionlessTerminalUpstreamReplyIsDelivered (0.03s)
    --- PASS: .../nxdomain_without_question_is_delivered (0.01s)
    --- PASS: .../refused_without_question_is_delivered (0.00s)
    --- PASS: .../servfail_without_question_is_delivered (0.00s)
    --- PASS: .../positive_answer_without_question_is_still_dropped (0.00s)
    --- PASS: .../nodata_without_question_is_still_dropped (0.01s)
ok  github.com/daeuniverse/dae/control
```

The test is not vacuous: with the new guard disabled, the three "delivered" subtests fail, each with `upstream … reply does not echo the request question (possible spoofing or upstream cross-talk); dropped` — exactly the regression this patch prevents.

### 3. Functional check on a real gateway

The same behaviour (inserted at the pre-rework call site, before the kdae rework moved it) has been running on an x86_64 OpenWrt router with real LAN clients and the real upstream, driven by the deterministic trigger from the Background section:

| check | before | after |
|---|---|---|
| LAN client `dig … in-addr.arpa PTR` | 5 s timeout | 15–17 ms `NXDOMAIN` |
| router log errors for that query family | +122 in 14.5 min | 0 (60 s and 3 min of natural traffic) |
| A / TXT / AAAA / NXDOMAIN regression | 15/15 | 15/15 |

Happy to repeat the functional run on this exact revision: I can deploy the build of `1c8803ba` to that router and re-run the same A/B if a maintainer wants it before labelling.

xunkutech:
Fixed in 727308c (kept as a single commit, as the contribution guide asks). The `unconvert` finding was a leftover `uint16(...)` cast around `dnsmessage.TypePTR` — the constant is already `uint16`.

Re-verified locally with the same linter version and the same flags as the workflow:

```
$ golangci-lint --version
golangci-lint has version 2.11.0 built with go1.26.1
$ golangci-lint config verify          # clean
$ golangci-lint run --timeout=10m --build-tags=dae_stub_ebpf ./control/... ./component/dns/...
0 issues.

$ go test -tags dae_stub_ebpf ./control/ -run TestQuestionlessTerminalUpstreamReplyIsDelivered -count=1
--- PASS: TestQuestionlessTerminalUpstreamReplyIsDelivered (0.04s)
ok  github.com/daeuniverse/dae/control
```

The runs for the new revision are again in `action_required` (first-time contributor), so they need an approval to start.

